PR #1, opened 2023-07-05
Introduce protobuf libraries

mara214:
Introduces:

- The `lib` package types (and thus adjusts the manifest format)
- Scoped and nested dependency resolution (installing libraries needed for api packages you consume)

### DIFF
--- a/src/package.rs
+++ b/src/package.rs
@@ -14,7 +14,10 @@ use serde::{Deserialize, Serialize};
 use tokio::fs;
 use walkdir::WalkDir;
 
-use crate::manifest::{self, Manifest, PackageManifest, RawManifest, MANIFEST_FILE};
+use crate::{
+    manifest::{self, Dependency, Manifest, PackageManifest, RawManifest, MANIFEST_FILE},
+    registry::Registry,
+};
 
 /// IO abstraction layer over local `buffrs` package store
 pub struct PackageStore;
@@ -22,7 +25,7 @@ pub struct PackageStore;
 impl PackageStore {
     /// Path to the proto directory
     pub const PROTO_PATH: &str = "proto";
-    /// Path to the api directory
+    /// Path to the lib directory
     pub const PROTO_LIB_PATH: &str = "proto/lib";
     /// Path to the api directory
     pub const PROTO_API_PATH: &str = "proto/api";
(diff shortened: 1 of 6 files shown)


src/package.rs
```diff
@@ -57,28 +65,75 @@ impl PackageStore {
...
+                '┗'
+            } else {
+                '┣'
+            };
+
+            tracing::info!(
+                "   {tree_char} installed {}@{}",
+                dependency.manifest.name,
+                dependency.manifest.version
+            );
```
wngr, line 135: If there's concurrent stuff happening the output might be intertwined. Maybe write it all at once?

src/package.rs
```diff
@@ -159,32 +250,94 @@ impl PackageStore {
...
 impl Package {
     /// Creates a new package
-    pub fn new(name: PackageId, version: String, tgz: Bytes) -> Self {
-        Self { name, version, tgz }
+    pub fn new(manifest: PackageManifest, tgz: Bytes) -> Self {
+        Self { manifest, tgz }
+    }
+
+    /// Decodes the tar and returns the encoded manifest
+    pub fn decode(tgz: Bytes) -> eyre::Result<Self> {
```
wngr, line 276: `impl TryFrom<Bytes>`?

wngr (commented):
Isolated changes LGTM. Could I trouble you for some tests :-)?

mara214:
> Isolated changes LGTM. Could I trouble you for some tests :-)?

Of course!

src/package.rs
mara214: there is no concurrent stuff happening, but i agree 😊 
i tried to use tracing spans but it complains about not using them across async ops

src/package.rs
mara214: true!

but the weird thing here is that we only do some introspection and then still store the compressed tgz

But i agree when seeing this a second time, it should be try from

src/package.rs
wngr: I think you can annotate the whole function or use a closure, see https://docs.rs/tracing/latest/tracing/struct.Span.html#in-asynchronous-code.

However this won't help here, as individual log events will be pushed to your subscriber anyway, just potentially augmented with some span metadata.
Spans might be interesting nonetheless get some timing information 🤷

mara214: Okay :) I'll fix the logging! Thx for the hint

PR #4, opened 2023-07-19
Implement code generation using tonic

mara214:
# Introduces

The `buffrs gen|generate --lang <lang>` command for the cli interface.

`buffrs::build(Language::Rust)` which installs all missing and outdated dependencies and then compiles all dependencies and outputs the generated code into `proto/build`

`buffrs::include!("<dependency-package-id>")` for easy usage of the generate code

# Example

In a buffrs project the cargo intgeration is now super simple!

Given a buffrs project which depends on the `demo` api package which contains only one proto file:

`proto/vendor/demo/demo.proto`: 
```proto
syntax = "proto3";

package demo;

message Demo {}
```

Building / installing the protocol buffers now only requires:

`build.rs`: 
```rust
fn main() {
    buffrs::build(buffrs::Language::Rust).unwrap();
}
```

If any dependencies are missing during the build process they will be automatically downloaded from the registry!

This means you can check out any buffrs + cargo project and `cargo run` will just work™

(And of course buffrs has to be added to the Cargo.toml as build dependency and direct dependency, and tonic + prost as normal dependencies)

---

The application code now only needs to do a `buffrs::include` to include the generated code

`src/main.rs`: 
```rust
mod protos {
    buffrs::include!("demo");
}

fn main() {
    dbg!(protos::demo::Demo {});
}
```


src/package.rs
```diff
@@ -140,7 +125,7 @@ impl PackageStore {
 
     /// Uninstalls a package from the local file system
     pub async fn uninstall(package: &PackageId) -> eyre::Result<()> {
-        let pkg_dir = Path::new(Self::PROTO_DEP_PATH).join(package.as_package_dir());
+        let pkg_dir = Path::new(Self::PROTO_VEDNOR_PATH).join(package.as_str());
```
cdbrkfxrpt:
```suggestion
        let pkg_dir = Path::new(Self::PROTO_VENDOR_PATH).join(package.as_str());
```

cdbrkfxrpt (changes requested):

Cargo.toml
```diff
@@ -23,17 +23,19 @@ color-eyre = "0.6"
 eyre = "0.6"
 flate2 = "1"
 futures = "0.3"
+git2 = "0.17"
 home = "0.5.5"
 human-panic = "1"
 keyring = "2"
-git2 = "0.17"
+protobuf-src = "1.1.0"
```
j-baker, line 30: I don't recommend usage of protobuf-src. I used to love it but IMO it's a false economy. It takes a long time to compile and because it has all the dependencies protobuf has, if it goes wrong it has very very hard to debug build issues. I think insofar as you want to ensure a specific protobuf version,it'd be better to check protoc's version deliberately.

src/generator.rs
```diff
@@ -0,0 +1,140 @@
...
+pub enum Generator {
+    /// The tonic + prost stack
+    Tonic,
+}
+
+impl Generator {
+    pub const TONIC_INCLUDE_FILE: &str = "mod.rs";
+
+    /// Run the generator for a dependency and output files into `out`
+    pub async fn run(&self, dependency: &Dependency, out: &Path) -> eyre::Result<()> {
```
j-baker, line 50: curiosity - wouldn't you want to pass in all dependencies at once? if dependencies reference each other won't Tonic want access to all of those types? Otherwise I'd at the very least expect clashing definitions of Any etc.

Cargo.toml
mara214: The compilation errors are catched by prost faik, i would really want to use a vendored thing, to keep the setup reproducible everywhere.

src/generator.rs
mara214: No! Buffrs sandboxes each dependency so if your deps look like:

```
you
  api-a@foo
    common@bar
  api-b@baz
    common@oof
```

You get:

```rust
mod api_a { mod common /* @bar for usage with api_a */ { .. } }
mod api_b { mod common /* @oof for usage with api_b */ { .. } }
```

src/generator.rs
```diff
@@ -0,0 +1,140 @@
...
+                fs::create_dir_all(&out)
+                    .await
+                    .wrap_err("failed to recreate dependency output directory")?;
+
+                let package = PackageStore::locate(&dependency.package);
+                let protos = PackageStore::collect(&package).await;
+
+                let includes = &[package];
+
+                tonic_build::configure()
```
j-baker, line 69: in this approach, I think you're pulling the tonic code-gen into the buffrs binary. The one concern I'd have here is dependency misalignment between buffrs and tonic - e.g. if if buffrs isn't designed to be run from a build script, what keeps tonic runtime dep + buffrs tonic-build dep in sync? Have you thought about this at all?

Cargo.toml
j-baker: What I mean is that protobuf-src runs the protobuf build. I had a bunch of issues where that build was simply failing with some cmake issue, which did not happen if I just compiled protobuf on its own, and this was then very hard to debug. Additionally, you won't be able to rely on protobuf-src for e.g. generating python or ts.

src/generator.rs
mara214: > buffrs isn't designed to be run from a build script

it is, either cli or `build.rs`

regarding the tonic version you use, we could either reexport tonic for the project, so it only has to use `buffrs::tonic as tonic` to keep them in sync, or trust the user to not mess up there and figure out the compatible versions

Cargo.toml
mara214: Ah! We could also use a vendored version which is precompiled

mara214: https://crates.io/crates/protoc-bin-vendored

src/generator.rs
j-baker: I can see why you've done this. I'm concerned that it leads to considerable ergonomic issues with regards to standard types - notably around conversions. For example, if APIs A and B both use some shared (and complex) type C, a consumer who reads a C from A and puts it into B will have to do conversions.
This seems to mitigate a large part of the advantage of being able to depend on shared protobuf libraries.

I'd wonder if there's some middle ground like Nix's 'follows' declarations? (you've used nix flakes a bunch, yeah?)

At the very least, it seems quite important to me that you don't use compile_well_known_types, so that at least Timestamp and Any don't have this issue.

src/generator.rs
j-baker: ack nice, loving the build.rs.

src/generator.rs
wngr: Upside is isolation, downside is that you can't use types of `api_a.common` with `api_b.common` w/o doing some conversion. I guess this a good tradeoff nevertheless though, but surprising coming from the status quo.

mara214: > This seems to mitigate a large part of the advantage of being able to depend on shared protobuf libraries.

I agree, this may also change in the future. For now users are required to implement type conversions, as the actual dependency resolution would be a too complex problem to tackle as a side project

This is not saying we wont do it, but rather, we cant time wise do it in a sensible way

mara214: I can disable the common type compilation 😊

Cargo.toml
wngr: Another alternative is to require `protoc` to be present on the host, and check its version range?

src/generator.rs
```diff
@@ -0,0 +1,140 @@
...
+/// mod protos {
+///     buffrs::include!("demo");
+/// }
+/// ```
+#[macro_export]
+macro_rules! include {
+    ($package:expr) => {
+        ::std::include!(concat!(
+            env!("CARGO_MANIFEST_DIR"),
+            "/proto/build/rust/",
```
wngr, line 135: Reuse `BUILD_DIRECTORY`

src/generator.rs
wngr: What about converting this to a macro so the dependency on `tonic_build` can be removed completely?

Also: building server/client could be made conditional.

wngr (commented):
🙌

src/generator.rs
mara214: I wanted!! But t is not possible with concat

mara214:
https://github.com/helsing-ai/buffrs/pull/4#discussion_r1274803878

see the discussion above, we probably use the vendored version, but not the src crate

src/generator.rs
mara214: Configuring buffrs build behavior is meant to go into the manifest `[build.rust]` section!

If this doesnt work for some reason, you can always opt out of the `buffrs::build` plug and play solution

Cargo.toml
mara214: I would opt against that if somehow possible

tetigi: After having seen the issues caused by `protoc-src` and others, I think that checking for `protoc` and validating its version-range is the lesser-evil here (where you can use the version check to ensure reproducibility). It doesn't "just-work" (which sucks), but I think the other options have much worse worst-cases :(

mara214: and using the vendored ones? @tetigi @wngr

PR #8, opened 2023-08-10
Migrate credential management away from keyring

mara214:
The keyring implementation sadly resulted in repromting users every time they installed packages with buffers. This pr is to follow cargo by creating a credentials.toml file in the home directory. It is less secure but also a huge win in terms of UX. 

Users can protect the file with according os permissions to gain security back.

### DIFF
--- a/src/credentials.rs
+++ b/src/credentials.rs
@@ -9,60 +9,60 @@ use crate::registry::ArtifactoryConfig;
 
 /// Global configuration directory for `buffrs`
 pub const BUFFRS_HOME: &str = ".buffrs";
-/// Filename of the configuration
-pub const CONFIG_FILE: &str = "config.toml";
+/// Filename of the credential store
+pub const CREDENTIALS_FILE: &str = "credentials.toml";
 
-/// Configuration format for storing authentication and settings
+/// Credential store for storing authentication data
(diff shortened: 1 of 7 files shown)


src/credentials.rs
```diff
@@ -9,60 +9,60 @@ use crate::registry::ArtifactoryConfig;
...
             .await
-            .wrap_err("Failed to write config")
+            .wrap_err("Failed to write credentials")
     }
 
-    /// Loads the configuration from the file system
+    /// Loads the credentials from the file system
     ///
-    /// Note: Initializes the configuration if its not present
+    /// Note: Initializes the credential file if its not present
```
ntrinquier, line 62:
```suggestion
    /// Note: Initializes the credential file if they are not present
```

src/credentials.rs
```diff
@@ -9,60 +9,60 @@ use crate::registry::ArtifactoryConfig;
...
     pub async fn load() -> eyre::Result<Self> {
         let Ok(cfg) = Self::read().await else {
-            let cfg = Config::default();
+            let cfg = Credentials::default();
```
ntrinquier, line 65:
```suggestion
            let credentials = Credentials::default();
```

PR #13, opened 2023-08-16
Refactor vendor dir

mara214:
## Changes

This pr refactors the package store from:

```
proto
  /vendor/
    <a>/
      Proto.toml
      vendor/<c>/..
    <b>/
      Proto.toml
      vendor/<c>/..
```

To 

```
proto
  /vendor/
    <a>/
      Proto.toml
      ..
    <b>/
      Proto.toml
      ..
    <c>/
      Proto.toml
      ..
```

And simplifies code generation implementation.

## Upsides

This drastically simplifies the package store layout, and more importantly, enables people to compile it using *any* protobuf code gen tool, such as protoc without any change.

Furthermore it removes the type duplication problem that occured when nesting dependencies:

Before the change, you had two version of the `c` library types:

```rust
mod a {
  mod c { /* ... */ }
}

mod b {
  mod c { /* ... */ }
}
```

Which is now resolved, since tonic generates this structure:

```rust
mod a { /* ... */ }
mod b { /* ... */ }
mod c { /* ... */ }
```

## Downsides

This change has the implication that *all* dependencies may only use one single version of `c` which is not ideal. We can and will solve this by introducing proper version resolution mechanism soon.

### DIFF
--- a/src/generator.rs
+++ b/src/generator.rs
@@ -45,7 +45,7 @@ impl Generator {
     pub const TONIC_INCLUDE_FILE: &str = "mod.rs";
 
     /// Run the generator for a dependency and output files into `out`
-    pub async fn run(&self, output: &Path) -> eyre::Result<()> {
+    pub async fn run(&self, output: impl AsRef<Path>) -> eyre::Result<()> {
         let protoc = protoc_bin_path().wrap_err("Unable to locate vendored protoc")?;
 
         std::env::set_var("PROTOC", protoc.clone());
--- a/src/generator.rs
+++ b/src/generator.rs
@@ -8,10 +8,7 @@ use protoc_bin_vendored::protoc_bin_path;
 use serde::{Deserialize, Serialize};
 use tokio::fs;
 
-use crate::{
-    manifest::Manifest,
-    package::{PackageId, PackageStore},
-};
+use crate::{manifest::Manifest, package::PackageStore};
 
 /// The directory used for the generated code
 pub const BUILD_DIRECTORY: &str = "proto/build";
@@ -37,14 +34,6 @@ impl fmt::Display for Language {
     }
 }
 
-/// A configuration passed to a generator to compile a package
-#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
-pub struct GenerationConfig<'i, 'p> {
-    pub package: &'i PackageId,
-    pub location: &'p Path,
-    pub out: &'p Path,
-}
-
 /// Backend used to generate code bindings
 #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub enum Generator {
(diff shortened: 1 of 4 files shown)


src/generator.rs
```diff
@@ -56,31 +45,23 @@ impl Generator {
...
-
-                let includes = &[&config.location];
-
                 tonic_build::configure()
                     .build_client(true)
                     .build_server(true)
                     .build_transport(true)
                     .compile_well_known_types(true)
-                    .out_dir(&out)
+                    .out_dir(output)
```
cdbrkfxrpt, line 64: if `output` is the output directory, and it is set via `.out_dir(...)`, i would probably simply name it `out_dir`

src/generator.rs
```diff
@@ -117,33 +98,18 @@ pub async fn generate(language: Language) -> eyre::Result<()> {
         out
     };
 
+    generator
+        .run(&out)
```
cdbrkfxrpt, line 102: rename `out` to `out_dir` for clarity; avoids the reader having to read the code block to determine what `out` refers to

src/package.rs
```diff
@@ -97,14 +95,29 @@ impl PackageStore {
...
         for (index, dependency) in dependencies.into_iter().enumerate() {
+            if let Ok(manifest) = Self::resolve(&dependency.package).await {
+                let existing = manifest.package.wrap_err(eyre::eyre!(
+                    "Found installed manifest for {} but it is malformed",
+                    dependency.package,
+                ))?;
+
+                eyre::ensure!(
+                    existing.version == dependency.manifest.version,
+                    "Another dependency in your project requires {}@{} which colides with {}@{} required by {}",
```
cdbrkfxrpt, line 109:
```suggestion
                    "A dependency of your project requires {}@{} which collides with {}@{} required by {}",
```

cdbrkfxrpt (approved):
a few very minor nits aside, this lgtm

src/generator.rs
```diff
@@ -56,31 +45,23 @@ impl Generator {
     pub const TONIC_INCLUDE_FILE: &str = "mod.rs";
 
     /// Run the generator for a dependency and output files into `out`
-    pub async fn run(&self, config: GenerationConfig<'_, '_>) -> eyre::Result<()> {
+    pub async fn run(&self, output: &Path) -> eyre::Result<()> {
```
wngr, line 48:
```suggestion
    pub async fn run(&self, output: impl AsRef<Path>) -> eyre::Result<()> {
```

src/generator.rs
```diff
@@ -56,31 +45,23 @@ impl Generator {
     pub const TONIC_INCLUDE_FILE: &str = "mod.rs";
 
     /// Run the generator for a dependency and output files into `out`
```
wngr, line 47:
```suggestion
    /// Run the generator for a dependency and output files at the provided path
```

wngr (approved):

PR #18, opened 2023-08-24
Start implementing semver

mara214:
# Changes

Use the `semver` crate's types `Version` and `VersionReq` instead of strings

# Implications

Existing version requirements in manifests will result in errors as the default operator is `Caret` and only `Exact` is supported as of right now.

Example:

```
[dependencies]
pkg = { version = "1.0.0", repository = "foo" }
```

Will have to be changed to:
```
[dependencies]
pkg = { version = "=1.0.0", repository = "foo" }
```

---

The proper evalutation / resolution of requirements will be implemented by the registry (see #12)

### DIFF
--- a/src/registry/artifactory.rs
+++ b/src/registry/artifactory.rs
@@ -40,6 +40,8 @@ impl Artifactory {
 impl Registry for Artifactory {
     /// Downloads a package from artifactory
     async fn download(&self, dependency: Dependency) -> eyre::Result<Package> {
+        todo!("IMPLEMENT VERSION RESOLUTION");
+
         let artifact_uri: Url = {
             let mut url = self.0.url.clone();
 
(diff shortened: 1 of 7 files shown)


src/registry/artifactory.rs
```diff
@@ -40,6 +40,36 @@ impl Artifactory {
...
+            .manifest
+            .version
+            .comparators
+            .first()
+            .wrap_err("internal error")?;
+
+        ensure!(
+            version.op == semver::Op::Exact && version.minor.is_some() && version.patch.is_some(),
+            "artifactory only support pinned dependencies"
+        );
```
Tpt, line 59: It might be nicer to move this validation where the version requirement is added. But I guess this will change soon so it's not a big deal.

Tpt (approved):

src/registry/artifactory.rs
mara214: Well, the requirements are valid in the manifest they just can't be used with the artifactory registry impl

PR #23, opened 2023-08-30
 Implement LocalRegistry for testing

uschi2000:


src/registry/test-data/test-api-0.1.0.tgz
mara214, line 1: Can we store the test data in the `tests/*` dirs eventually?

mara214 (commented):
Few nits but generally it lgtm!

src/registry/test-data/test-api-0.1.0.tgz
uschi2000: done

src/registry/mod.rs
```diff
@@ -3,6 +3,8 @@
 use crate::{manifest::Dependency, package::Package};
 
 mod artifactory;
+#[cfg(test)]
```
uschi2000, line 6: fyi @mara214 , I am now hiding the whole module in the test config

mara214 (approved):

PR #24, opened 2023-08-30
Plumb through a base_dir everywhere to facilitate composition and testing; add first store test

uschi2000:


uschi2000:
Not super happy yet with the overall decomp, but I think this goes in the right direction:
- all methods that read or write from disk are now relative to a base directory
- in the CLI, the base directory is `.` (ie, yielding the same behavior as before)
- in tests, we can set a custom base_dir

src/registry/local.rs
```diff
@@ -90,53 +90,45 @@ impl Registry for LocalRegistry {
...
+        let tmp_dir = env::temp_dir();
+        let registry = LocalRegistry::new(tmp_dir.clone());
+
+        // Create an (in-memory) release of the test proto package
+        let package_store = PackageStore {
+            base_dir: PathBuf::from("tests/data/test-api"),
         };
-        let package_bytes =
-            Bytes::from(include_bytes!("../../tests/data/test-api-0.1.0.tgz").to_vec());
+        let release = package_store.release().await.unwrap();
```
uschi2000, line 112: this simplification is possible because we can now call `release` to create a release, rather than having to check in the pre-tarred tgz

src/manifest.rs
```diff
@@ -1,13 +1,17 @@
...
+
 use eyre::Context;
 use semver::{Version, VersionReq};
 use serde::{Deserialize, Serialize};
-use std::{collections::HashMap, fmt};
 use tokio::fs;
 
 use crate::package::{PackageId, PackageType};
 
+// TODO(rfink): For symmetry between the vendored and the local packages, we should consider
```
uschi2000, line 13: worth discussion, @mara214 . I find the asymmetry between the local and the vendored packages a bit strange

uschi2000:
Main thing that's still a bit weird is that we create a bunch of PackageStore instances ad hoc in a few places. I would suggest tackling this in a follow-up, because this is already quite large.

mara214:
I honestly think we should not make the package store object oriented and use snapbox for testing that does IO similar to cargo. Unit tests should never do stateful things / be dependent on the system state (its an antipattern in my personal opinion)

The testing plan for the cli was to mimic cargos testing setup and use unit test only for pure types / areas of the project

mara214:
Imo testability should not degrade readability / add complexity 

Have you had a look at cargos testsuite yet?

uschi2000:
I haven't look at the cargo tests, no. snapbox isn't particular approach, the documentation examples are not self-contained and snapbox itself is practically untested... maybe a bit strange for a testing framework?

Ultimately, I think you need to decide how you would like to test buffrs in this repo, I primarily care that we have non-trivial test coverage. Personally, I find code a lot more easy to (re-)use (and test!) when it doesn't make deep assumptions about environment concerns such as env vars or working directory.

mara214:
I agree with your testability concerns but removing those assumptions does add the complexity im trying to keep as low as possible

I personally would want to add an E2E testsuite which asserts the cli (thus compiled binary) stdout / stderr output and modifications to the filesystem and keep internals as flexible as possible for now as the code base has not stabilized – does that approach sound feasible to you?

---

Snapbox is not a fixed choice, there are a lot of similar libraries like assert_cmd + predicates

See https://rust-cli.github.io/book/tutorial/testing.html for reference

mara214:
The guide I referenced basically splits testing into "business logic" and "fs/io" tests, the former are implemented as unit tests and the later as e2e tests

uschi2000:
> I agree with your testability concerns but removing those assumptions does add the complexity im trying to keep as low as possible
> 
> I personally would want to add an E2E testsuite which asserts the cli (thus compiled binary) stdout / stderr output and modifications to the filesystem and keep internals as flexible as possible for now as the code base has not stabilized – does that approach sound feasible to you?
> 
> Snapbox is not a fixed choice, there are a lot of similar libraries like assert_cmd + predicates
> 
> See https://rust-cli.github.io/book/tutorial/testing.html for reference

If the testing framework is good, then yes, this is of course a feasible setup. As a litmus test, I would hope that simple tests are <10LOC.

src/package.rs
```diff
@@ -437,3 +450,65 @@ impl fmt::Debug for PackageId {
             .finish()
     }
 }
+
+#[cfg(test)]
+mod tests {
```
mara214, line 455: I think we should definitely keep this test, but i would prefer to move it to the `tests` directory as it does IO 🙂

Can we get around the oop `PackageStore` and `Cmd` structs for now and keep the testing aspect?

I do see value in being able to configure the directory eg. when you want to run commands in a nested dir / need to bubble up to the next manifest as in cargo workspaces but i think this feature is not so pressing right now to be worth the required complexity

uschi2000:
Won't do.

PR #32, opened 2023-09-08
Send the X-JFrog-Art-Api instead of using basic auth

kihehs:
Using this header means we don't have to send the username anymore. I believe this means we can remove usernames completely.


mara214:
Hi @kihehs, is this strictly backwards compatible / tested that it doesn't break anything? 😊 

Also a good follow up would be to remove the username from the `Credentials` file

kihehs:
I can't test publishing and downloading right now, but it would appear that the tests are green, for what it's worth :)

kihehs:
Okay, now I know that downloading works, so presumably publishing does as well :)

mara214:
The docs should be updated before we can merge, but other than that it lgtm!

https://github.com/helsing-ai/buffrs/blob/main/docs/src/getting-started/installation.md

kihehs:
Alright, I removed the username line, and did a cursory search. I think it's finally gone :)

mara214 (approved):

PR #34, opened 2023-09-08
Import registry ui template

mara214:


mara214:
nvm @cdbrkfxrpt, not much to review yet i guess

PR #36, opened 2023-09-08
Add E2E tests where possible

mara214:


tests/cmd/login/mod.rs
```diff
@@ -0,0 +1,22 @@
+use crate::VirtualFileSystem;
+
+#[test]
+fn fixture() {
+    let vfs = VirtualFileSystem::empty().with_virtual_home();
+
+    crate::cli!()
+        .arg("login")
+        .arg("--url")
```
uschi2000, line 9: not sure if possible, but this would be more legible:
```
crate::cli!()
        .arg("login")
        .args("--url https://org.jfrog.io/artifactory")
        .args("--username foo")
```

src/main.rs
```diff
@@ -221,17 +221,15 @@ mod cmd {
         let dependency = manifest
             .dependencies
             .iter()
-            .find(|d| d.package != package)
+            .find(|d| d.package == package)
```
uschi2000, line 224: this was just a straight up bug?

uschi2000 (approved):

src/main.rs
mara214: Yep!

tests/cmd/login/mod.rs
mara214: not possible :/ one would have to pass a slice to `args` which is not really better imo

Issue #3, opened 2023-07-12
 lockfile


j-baker:
Internally am trying to incrementally ensure our ecosystem is as nix-buildable as possible. One of the constraints imposed by Nix is that derivations should be immutable - every call must return the same value. This is valuable for obvious reasons, and Nix reasonably doesn't trust that URLs can't change.

When integrating an external package manager (like Cargo, or buffrs) into Nix, this typically means that the integration occurs along the lines of the package manager's lockfile - the config+lockfile together contain enough information to reconstruct a URL and a cryptographic hash of that content.

It would be convenient if buffrs utilised a similar lockfile, which could be e.g. updated at generation time like Cargo.lock. This would have the additional benefit of detecting any accidental overwrites, and while internally we'll usually use Artifactory with a config which disallows these, I'd expect that the mooted S3 integration would benefit from it.

mara214:
I agree! Thanks for suggesting this! 😊 
I'll start to implement this this week or next week

mara214:
We should also store the buffrs / protoc version in the lockfile to detect problems in codegen early on!

j-baker:
Is there an ETA on this?

mara214:
Starting September we have additional resources on buffrs to tackle this so I assume 1-3 weeks - is that okay or do you need it earlier?

mara-schulke:
#38 is drafting the implementation for this 😊 

LMK if you have ideas regarding the implementation / feel free to take a look at the pr 

--

Is there any special contract the lockfile has to fulfill to be meaningful for the nix integration?

So far we pin these values for each installed package:

- Version
- Name
- Artifactory Repository
- Checksum
- Dependencies of the Package

mara-schulke:
Lockfiles landed @j-baker!

Issue #26, opened 2023-09-04
Add meaningful unit tests


mara214:
We should test eg. parsing / deserialization logic + general business logic which does not rely on IO

asmello:
Note echoed from #38 - we need to mock registry interactions so we can test local logic independently of it.

xfbs:
I think this is something I am interested in tackling.

mara-schulke:
Solved

PR #38, opened 2023-09-10
Implement Lockfiles (`Proto.lock`) 

mara-schulke:
# Introduces

- [x] A `Proto.lock` (see #3)

# Breaks

- Existing manifests without `[package]`
- Existing manifests without `[dependencies]`


src/lock.rs
```diff
@@ -0,0 +1,166 @@
...
+                .manifest
+                .dependencies
+                .iter()
+                .cloned()
+                .map(|d| d.package)
+                .collect(),
+        }
+    }
+
+    pub fn validate(&self, other: &Self) -> eyre::Result<()> {
```
mara-schulke, line 46: Can you add doc comments to everything that is public?

src/manifest.rs
```diff
@@ -16,8 +16,8 @@ pub const MANIFEST_FILE: &str = "Proto.toml";
 /// empty fields.
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 pub struct RawManifest {
```
mara-schulke, line 18: can we make this pub(crate)? not done by you initially but just recognizing it should be

src/package.rs
```diff
@@ -370,87 +352,225 @@ impl fmt::Display for PackageType {
...
+impl fmt::Debug for PackageName {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        f.debug_tuple("PackageId")
+        f.debug_tuple("PackageName")
             .field(&format!("{self}"))
             .finish()
     }
 }
+
+pub struct ResolvedDependency {
```
mara-schulke, line 446: also doc comments

src/command.rs
```diff
@@ -0,0 +1,305 @@
...
+        name: &PackageName,
+        tree: &DependencyTree,
+        locked: &mut Vec<LockedPackage>,
+        prefix: String,
+    ) -> eyre::Result<()> {
+        let resolved = tree
+            .get(name)
+            .wrap_err("Dependency missing from dependency tree")?;
+
+        PackageStore::unpack(&resolved.package).await?;
```
mara-schulke, line 210: ah!

src/manifest.rs
asmello: That makes sense, but I think it's even better to implement a `Manifest::from_string()` function so we don't need to expose this struct outside its module (similar to how I implemented `RawLockfile`).

src/package.rs
```diff
@@ -62,83 +66,32 @@ impl PackageStore {
...
+            package.manifest.package.name,
+            package.manifest.package.version,
             pkg_dir.display()
         );
 
         Ok(())
     }
 
-    /// Installs a package and all of its dependency into the local filesystem
-    pub async fn install<R: Registry>(dependency: Dependency, registry: R) -> eyre::Result<()> {
```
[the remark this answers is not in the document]
asmello: The logic got refactor into `DependencyTree` and `command::install()`. All that `PackageStore` is used for now is `PackageStore::unpack()` after all downloads and validations have taken place.

src/manifest.rs
asmello: Made the raw struct private (and implemented a few to/from string traits to support converting the manifest to/from TOML).

src/lock.rs
asmello: Think I covered everything now (🤞 ).

docs/src/faq.md
```diff
@@ -0,0 +1,27 @@
...
+```
+
+Since the CLI won't allow you to publish a package with an invalid manifest,
+you may wonder why some published packages may be broken in this way. The
+reason is that all fields used to be optional -- `impl` packages didn't need a
+`package` section, and `lib` packages didn't require `dependencies`. When the
+lockfile functionality was first introduced, both sections were made mandatory
+for all package types, which broke parsing of older manifests.
+
+To fix this you must depend on a newer version of the dependency that publishes
```
j-baker, line 26: have you thought about storing a version in the buffrs manifest so that you can indicate incompatibility via something other than failure to deserialise?

src/command.rs
```diff
@@ -0,0 +1,306 @@
...
+/// Publishs the api package to the registry
+pub async fn publish(
+    credentials: Credentials,
+    repository: String,
+    allow_dirty: bool,
+    dry_run: bool,
+) -> eyre::Result<()> {
+    if let Ok(repository) = git2::Repository::discover(Path::new(".")) {
+        let statuses = repository
+            .statuses(None)
```
j-baker, line 139: I wrote the same code yesterday and irrc this doesn't filter out ignored files - rather you need to do `Some(Default::default())`. Did you test this?

Likewise - why does this matter to buffrs? Surely for buffrs there are like 2 files that matter and who cares if the user has a bunch of other files edited - that's not buffrs' purpose.

src/command.rs
```diff
@@ -0,0 +1,306 @@
...
+                .for_each(|s| tracing::error!("{}", s.path().unwrap_or_default()));
+
+            tracing::error!("\nTo proceed with publishing despite the uncommitted changes, pass the `--allow-dirty` flag\n");
+
+            eyre::bail!("Unable to publish a dirty git repository");
+        }
+    }
+
+    let artifactory = {
+        let Some(artifactory) = credentials.artifactory else {
```
j-baker, line 156: this might be more easily written as

```rust
let Some(artifactory) = credentials.artifactory else {
  bail!("Unable to publish package to artifactory, please login using 'buffrs login'");
};
let artifactory = Artifactory::from(artifactory);
```

src/command.rs
```diff
@@ -0,0 +1,306 @@
...
+
+    artifactory.publish(package, repository).await?;
+
+    Ok(())
+}
+
+/// Installs dependencies
+pub async fn install(credentials: Credentials) -> eyre::Result<()> {
+    let artifactory = {
+        let Some(artifactory) = credentials.artifactory else {
```
j-baker, line 182: credentials here should be optional - e.g. Artifactory has anonymous mode auth

src/command.rs
```diff
@@ -0,0 +1,306 @@
...
+        .wrap_err_with(|| format!("Failed to generate language bindings for {language}"))?;
+
+    Ok(())
+}
+
+/// Logs you in for a registry
+pub async fn login(mut credentials: Credentials, url: url::Url) -> eyre::Result<()> {
+    let mut cfg = ArtifactoryConfig::new(url)?;
+
+    let password = {
```
j-baker, line 273: you might want to use a library which avoids putting the token in stdout as it's typed in?

src/generator.rs
```diff
@@ -1,3 +1,5 @@
+// (c) Copyright 2023 Helsing GmbH. All rights reserved.
```
j-baker, line 1: I don't think this is the right header for a project that's licensed under Apache 2.0. W/ apache 2.0, all files need an apache 2 header, see manual https://github.com/helsing-ai/buffrs/blob/main/LICENSE#L179

src/lock.rs
```diff
@@ -0,0 +1,174 @@
...
+            Lockfile::read().await
+        } else {
+            Ok(Lockfile::default())
+        }
+    }
+
+    /// Persists a Lockfile to the filesystem
+    pub async fn write(&self) -> eyre::Result<()> {
+        let raw = RawLockfile {
+            packages: self
```
j-baker, line 147: it'd be nice here for any lockfile to be _canonical_ - that there's only one valid written down version of a lockfile. This ensures that random changes in implementation don't lead to massive diffs.

tests/cmd/package/out/Proto.toml
```diff
@@ -2,3 +2,5 @@
 type = "lib"
 name = "lib"
 version = "0.0.1"
+
+[dependencies]
```
j-baker, line 6: it'd be nice to have some tests here which cover lockfiles

src/lock.rs
```diff
@@ -0,0 +1,174 @@
...
+struct RawLockfile {
+    packages: Vec<LockedPackage>,
+}
+
+/// Captures metadata about currently installed Packages
+///
+/// Used to ensure future installations will deterministically select the exact same packages.
+#[derive(Default)]
+pub struct Lockfile {
+    packages: HashMap<PackageName, Arc<LockedPackage>>,
```
j-baker, line 113: why the Arc here? doesn't feel like this is a place which benefits from the sharing - can't you just return an `&LockedPackage` or just clone it?

src/lock.rs
```diff
@@ -0,0 +1,174 @@
...
+/// Captures immutable metadata about a given package
+///
+/// It is used to ensure that future installations will use the exact same dependencies.
+#[derive(Serialize, Deserialize, Clone)]
+pub struct LockedPackage {
+    pub name: PackageName,
+    pub checksum: String,
+    pub repository: String,
+    pub version: Version,
+    pub dependencies: Vec<PackageName>,
```
j-baker, line 27: It's possible I've misunderstood how buffrs has evolved. As I've understood it, Buffrs doesn't do dependency resolution - rather it recompiles and vendors packages - such that you get some package duplication.

This would imply that dependencies would need to depend on both PackageName and Version. Is this still how buffrs works?

src/lock.rs
```diff
@@ -0,0 +1,174 @@
...
+
+pub const LOCKFILE: &str = "Proto.lock";
+
+/// Captures immutable metadata about a given package
+///
+/// It is used to ensure that future installations will use the exact same dependencies.
+#[derive(Serialize, Deserialize, Clone)]
+pub struct LockedPackage {
+    pub name: PackageName,
+    pub checksum: String,
```
j-baker, line 24: ideally we could put some stronger types around this? e.g. at this point, are we able to describe more about what hte checksum is? e.g. a Sha256Sum? Is the repository an HttpUrl?

j-baker: is the checksum specified as lower camel or upper camel?

j-baker: also - nit - checksum generally refers to values designed for purposes of detecting bit flip errors - more commonly in these lockfiles the value is called 'hash' or similar. Ideally if you know it's a sha256 hash, it's 'sha256_hash'.

src/command.rs
```diff
@@ -0,0 +1,306 @@
...
+        .parse::<VersionReq>()
+        .wrap_err_with(|| format!("Invalid version requirement supplied: {package}"))?;
+
+    let mut manifest = Manifest::read().await?;
+
+    manifest
+        .dependencies
+        .push(Dependency::new(repository.to_owned(), package, version));
+
+    manifest.write().await
```
j-baker, line 89: should add and remove also be updating the lockfile?

src/command.rs
```diff
@@ -0,0 +1,306 @@
...
+    manifest.write().await?;
+
+    PackageStore::create().await
+}
+
+/// Adds a dependency to this project
+pub async fn add(dependency: String) -> eyre::Result<()> {
+    let lower_kebab = |c: char| (c.is_lowercase() && c.is_ascii_alphabetic()) || c == '-';
+
+    let (repository, dependency) = dependency
```
j-baker, line 61: I'd recommend here separating out the type 'DependencyString' from the thing that adds it to the manifest. Why? because you want to write tests.

src/lock.rs
```diff
@@ -0,0 +1,174 @@
...
+                    }],
+                },
+                repository: self.repository.clone(),
+            },
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+struct RawLockfile {
```
j-baker, line 104: I would be highly tempted to put a hash of the manifest into the lockfile, as Poetry does. Why? It means you can trivially detect whether the lockfile is out of date with the dependencies.

src/command.rs
```diff
@@ -0,0 +1,306 @@
...
+
+    if !dry_run {
+        std::fs::write(path, package.tgz).wrap_err("failed to write package to filesystem")?;
+    }
+
+    Ok(())
+}
+
+/// Publishs the api package to the registry
+pub async fn publish(
```
j-baker, line 131: I would be highly tempted to put the version in the dependency only at publish time. It's not atypical and avoids the issue we end up having a lot where we leave repos dirty post publish.

docs/src/faq.md
asmello: @mara-schulke what do you think? I'm in favour of this, though we probably won't start using it seriously until 1.0.

asmello:
@j-baker the `command.rs` file is actually mostly pre-existing code, which I moved to a separate file so it could be reused by the binary and library targets. While we should follow up on the feedback you have for that module, for the sake of keeping this PR centred around Lockfile related changes I'll mark those comments as resolved and create a separate issue to  follow up.

src/command.rs
asmello: I thought about that, and I think yes, ideally. But since we're already doing substantive changes here I'm in favour of following up on that in a later PR.
@mara-schulke FYSA

asmello: Created #60 to track. Resolving for now (but happy to unresolve if Mara thinks this is worth scoping now).

src/command.rs
asmello: Can you elaborate on this? I don't follow how the repository may end up dirty post publish.

src/command.rs
asmello: Tracked in #61

src/command.rs
asmello: Tracked in #59

src/command.rs
asmello: #62

src/command.rs
asmello: I think #56 covers this.

src/command.rs
asmello: #63

src/command.rs
j-baker: I meant the version in the manifest. It’s typical to leave versions in manifests permanently at 0.0.0 or whatever and then when one publishes, they first rewrite the manifest, and then publish. It’d be nice if buffrs got the version out of the manifest, only adding it at publish time.

src/generator.rs
asmello: Good flag. I copied this from another file but I agree it doesn't seem right. Will fix tomorrow and we should probably follow up with a fix to other files too.

src/lock.rs
asmello: I actually tend to agree with all points. Unless @mara-schulke opposes I will change this tomorrow.

src/lock.rs
asmello: Ah, currently we only support exact versions, so if a dependency is required by two packages in the graph it will error out unless both request the same version. We'll need to make some changes to the `DependencyGraph` algorithm later to handle actual version resolution.

docs/src/faq.md
mara-schulke: We are storing a version in the manifest?

src/lock.rs
mara-schulke: No - buffrs does dependency resolution! We introduced this in v0.4

asmello: I'd actually value your input on whether this is headed in the right direction for that. My naive resolution algorithm would look somewhat like this:
1. When a dependency is first required (eg, it is included in the manifest), we assume the latest allowed version and recurse its sub-dependencies at that version.
2. If a sub-dependency (or another dependency) later is found to require the same package again, we add a constraint, recalculate the latest allowed version and schedule the sub-dependencies for traversal again (if they changed).

There might be edge cases leading to cycles which we'd need to detect, but I think this should work in general. I could be wrong... Haven't thought too deeply about it yet, but if I'm right the current data structure should work as is.

src/lock.rs
asmello: Hmmm that's not necessarily true though, at least the way we're using the lockfile. Even if you add or remove some dependencies, it doesn't entirely invalidate the lockfile. The unchanged dependencies can be reused as they are. Same goes if you change the version requirement and they still resolve to the same versions.

I think we're copying the model Cargo uses, unless I've misunderstood something (quite possible).

src/lock.rs
j-baker: Then is it worth leaving room for the dependencies in the lockfile to be dependency@versions? Rather than a single version thing? Or is it intended that buffrs unifies into a single tree permanently?

src/lock.rs
asmello: This is a bit residual, originally I was going to implement construction of the dependency graph in parallel and this was a good way to share the locked packages across tasks without cloning. Now it's still helpful in places like https://github.com/helsing-ai/buffrs/blob/mara/lockfile/src/package.rs#L548 but it's not very important. I don't feel strongly against removing it, but I do want to eventually improve concurrency, so it might become relevant again. @mara-schulke what's your take?

src/lock.rs
asmello: Oh yeah, good point. I was going to sort stuff to make it deterministic but forgot about it. Will fix tomorrow.

tests/cmd/package/out/Proto.toml
asmello: That's fair, but unfortunately that'd require registry interactions which we're currently unable to mock. I think #26 and #30 partially cover that, but I'll add a note about this specifically.

docs/src/faq.md
asmello: I think James was talking about storing a separate "manifest specification version" (similar to how cargo has `edition` and protobuf has the syntax version).

src/command.rs
asmello: I'm not sure this is meant to be the role of the package manager. We've been mimicking Cargo's design and it doesn't do anything special about the package version, it just reads it from the manifest. Updating the lockfile is one thing, but messing with the manifest behind your back sounds iffy.

mara-schulke: I think the "we use 0.0.0" as placeholder is something that is user specific / out of scope

docs/src/faq.md
asmello: Regardless I think this is out of scope for this PR. We can create a separate issue to track.

j-baker: Yes - it's typical to store a 'manifest version' in the manifest so compatibility can be flagged if breaking changes are made. This avoids semantic misinterpretation, making it easy to ahead-of-time surface "buffrs is out of date" rather than surfacing "some field is missing".

e.g. store

```
manifest-version: "1.2"
```

and if buffrs reads a manifest it can't interpret (because it uses some incompatible feature) you can error that the manifest is out of date.

src/lock.rs
j-baker: note - would be good to sort internally (e.g. on list of dependencies) as well as global list.

src/lock.rs
j-baker: My sense is that for what you're trying to do here, you care about _concurrency_ but not _parallelism_ - that's to say - you want to kick off your requests to artifactory in parallel but you don't care about running your whole algorithm in parallel.

This is because the dependency resolution algorithm should be defined to be cheaply implementable.

src/lock.rs
j-baker: I didn't see andre's reply.

My understanding of how you do simple dependency resolution in a statically typed language (which this is!) is:

1. You always pick the max version of what you've seen.
2. You always use that version, user gets no choice over the matter.

And if there's breakage, there's breakage and dependencies can't be used together. The alternatives are:

1. Do what python does - allow the user to specify dependency ranges, carets, tildes, all kinds of crap. The downside of this is that it's easy for one package to specify incompatibility anyway and so you end up with deps that can't be used together all the time even though the code works fine. This is good for Python as Python is fundamentally an untyped language and so you need the metadata.
2. Do what Rust does - allow multiple crates to be in use at any one time.
3. Do what Java does - if you make lots of breaking changes, that's your bad, don't do that.

What Python does is basically unsuitable for buffrs and unnecessary. The difference is whether you think you'll go down the java path or the rust path. If you might go down the rust path, probably worth containing the version in the lockfile spec.

An additional comment here: there might be no need to actually store the dependencies in the lockfile entry. In Python or Rust it _can_ be useful because you can put your feature flags and whatnot into the lockfile so you can filter down what you need to download based on platform and configuration. Whereas buffrs doesn't really have this. Without feature flags, platform specific considerations etc, you might just want to store the list of dependencies.

j-baker: Last comment on lockfile structure, but i promise, this one is important! At Palantir, we did java lockfiles https://github.com/palantir/gradle-consistent-versions/blob/develop/versions.lock - and a thing we noticed was that with a naive lockfile, it's easy to end up with semantic merge conflicts which cause the lockfile to be broken without generating a git conflict.

Effectively, you want unrelated concurrent updates to not generate git conflicts, but for related concurrent updates to generate git conflicts.

A simple (slightly contrived, but I promise this comes up in the real world quite a bit - I just can't remember the exact case) example of this would be:

```
x@1.1 depends on a@1.1 which depends on c@1.4
x@1.1 depends on b@1.2 which depends on c@1.4
a@1.2 removes the dependency on c
b@1.3 removes the dependency on c
```

and now if I had concurrent PRs to update `a` and `b`, neither of my PRs would remove the dependency on `c` from the lockfile - but after they merge, `c` should no longer be in the lockfile.

At palantir the goal was to have a very compact lockfile, with the hope being that instead of reviewing the list of dependency updates, you'd review the lockfile to see what had changed. So there the solution was that you'd build a hash of your dependents and stick that on the line.

So in the above case, the c dependency would have hash(a, b) next to it, and a change to either a or b would change the line referring to c.

j-baker: the tl;dr of the above being

- maybe discard the dependency graph from the lockfile?
- maybe instead put a hash of it on the consumers? not useful for dependency resolution but useful for causing merge conflicts

src/lock.rs
j-baker: after further thought, i disagree with this suggestion now beacuse it unnecessarily generates git merge conflicts. However, the thing I was pointing out was not relating to entirely invalidating - rather that it makes it easier to check whether the lockfile is out of date and therefore in need of recomputing. It's not about it being partially valid per se.

src/command.rs
j-baker: Out of scope in terms of what, exactly, Mara? Cargo made that design choice because it means that a crate is just a tar file - you can process a git, source, registry dependency in exactly the same way. Same thing for Python, for the same reason. But it's inconvenient for anyone using git versions, which are relatively commonly used - it means you either have a release step where you update the versions in your repo and commit that or you have an overwriting step at publish time. Avoiding needing to write down your own version in your manifest avoids that extra step.

Put another way:
- buffrs publishing as done by us internally seemingly requires overwriting of the buffrs manifest pre-publish to make sure the versions line up.
- buffrs also blows up if you have a dirty repo at publish time

How do we work around this? Doing a temporary commit of the buffrs manifest pre-push?

src/lock.rs
asmello: Yes, this is what I had in mind, but if you want tasks to have concurrent access to the lockfile (even read-only) you need to either clone it or wrap it in an Arc. That said I think I'll just leave that to the future and keep things simpler for now.

src/lock.rs
asmello: Ack, thanks for the clarification. That makes sense.

docs/src/faq.md
asmello: Created #64 to track / for further discussion.

mara-schulke: Agreed, we can think about this! But its to early to define a stable version for the manifest though

src/lock.rs
mara-schulke: I also agree

mara-schulke:
FYI @asmello i merged @kihehs pr so we got a couple of conflicts

src/lock.rs
asmello: Done. I actually used the term `digest` instead as that's a bit more formal / less ambiguous. Also changed how the hash is serialised so it encodes the hash algorithm, which helps with validation and may come in handy if we ever change the algorithm.

asmello:
> FYI @asmello i merged @kihehs pr so we got a couple of conflicts

"a couple"... 💦

src/command.rs
asmello: I think the way it's supposed to work is that you tie publishing to releases. So as a preparatory step you add a (permanent) commit that bumps the package version, then you tag that commit with a matching (eg, GitHub) release, and CI takes care of publishing the release (or alternatively you do it locally).

I get where you are coming from though. Keeping the manifest and repository versions in sync is riddled with pitfalls and poor UX. One generally needs sophisticated scripting just to glue those together with minimal risk of mistakes, though nothing stops someone from bypassing that. Ideally versions would be unified so bumps are transactional and there's no chance of desync between repository and registry.

It's not just Cargo and PIP that work this way, btw. Every single package manager that I know has the same issue. That doesn't mean we can't take a different approach and integrate with git / version control, but since that diverges from the beaten path it'd need thoughtful planning.

Let's move this discussion to its own space, as it's another instance of something that doesn't directly tie to the Lockfile functionality.

Tracking in #68

src/lock.rs
asmello: > maybe discard the dependency graph from the lockfile?

Rationale for keeping it in is that it could be useful for local diagnostics. Without that information we'd need to use the registry to resolve the dependency tree for inspection (for some `explain` command or similar).

> maybe instead put a hash of it on the consumers? not useful for dependency resolution but useful for causing merge conflicts

I think we may get by with just a count of dependants. I'm introducing a lockfile version field so that we can handle future migrations if necessary.

asmello: Ok, implemented the dependants count.

alexespencer (approved):
Nice! Edit: this was meant to be against a single line, trying to delete...

src/lock.rs
```diff
@@ -0,0 +1,328 @@
...
+    }
+}
+
+impl PartialEq for Digest {
+    fn eq(&self, other: &Self) -> bool {
+        self.algorithm == other.algorithm && self.bytes == other.bytes
+    }
+}
+
+impl Eq for Digest {}
```
jonhoo, line 141: I think these can be derived.

src/lock.rs
```diff
@@ -0,0 +1,328 @@
...
+
+impl Ord for Digest {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.bytes.cmp(&other.bytes)
```
jonhoo, line 145: Why doesn't this have to take into account the algorithm?

src/lock.rs
```diff
@@ -0,0 +1,328 @@
...
+    /// Captures the source, version and checksum of a Package for use in reproducible installs
+    pub fn lock(
+        package: &Package,
+        registry: RegistryUri,
+        repository: String,
+        dependants: usize,
+    ) -> Self {
+        let digest = digest::digest(&digest::SHA256, &package.tgz)
+            .try_into()
+            .unwrap(); // won't fail as SHA256 is sure to be supported
```
jonhoo, line 179: I would use `.expect` here with that message.

src/lock.rs
```diff
@@ -0,0 +1,328 @@
...
+            registry,
+            repository,
+            digest,
+            version: package.version().to_owned(),
+            dependencies: package
+                .manifest
+                .dependencies
+                .iter()
+                .cloned()
+                .map(|d| d.package)
```
jonhoo, line 192: Recommend mapping before cloning so you don't clone all the fields.

src/lock.rs
```diff
@@ -0,0 +1,328 @@
...
+            "Version mismatch - expected {}, actual {}",
+            self.version,
+            package.version()
+        );
+
+        Ok(())
+    }
+
+    /// Constructs a Dependency instance with matching metadata
+    pub fn as_dependency(&self) -> Dependency {
```
jonhoo, line 227: I think you can only implement this while pinned versions are the only dependency type supported. If you ever anticipate to support more, I would avoid including this `fn` since it _will_ go away in the future.

src/lock.rs
```diff
@@ -0,0 +1,328 @@
...
+                },
+                registry: self.registry.clone(),
+                repository: self.repository.clone(),
+            },
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+struct RawLockfile {
```
jonhoo, line 248: Given that this is versioned, I would probably write it as an `enum` straight away so that you don't need to change lots of code later when you _do_ cut a new version.

FYI, you'll probably want to write your own deserializer to switch on `version` is my guess.

jonhoo: You don't _have_ to do it in this PR (it's already quite large), but probably worth doing proactively.

Cargo.toml
```diff
@@ -35,10 +35,12 @@ eyre = "0.6"
 flate2 = "1"
 futures = "0.3"
 git2 = { version = "0.17", optional = true }
+hex = "0.4"
```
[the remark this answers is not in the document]
asmello: Hmm I assume this is what cargo produced (Mara added this before I picked up the branch). I think the lockfile should prevent issues but I'll pin down the patch as there's no real reason not to.

docs/src/faq.md
```diff
@@ -1,5 +1,31 @@
...
+   2:   | ^
+   2: missing field `dependencies`
+```
+
+Since the CLI won't allow you to publish a package with an invalid manifest,
+you may wonder why some published packages may be broken in this way. The
+reason is that all fields used to be optional -- `impl` packages didn't need a
+`package` section, and `lib` packages didn't require `dependencies`. When the
+lockfile functionality was first introduced, both sections were made mandatory
+for all package types, which broke parsing of older manifests.
```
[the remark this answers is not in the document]
asmello: Yes, that's right. I assume this is just to match Cargo — or was there another reason @mara-schulke?

src/manifest.rs
```diff
@@ -99,13 +121,29 @@ impl From<RawManifest> for Manifest {
...
+    fn try_into(self) -> Result<String, Self::Error> {
+        toml::to_string_pretty(&RawManifest::from(self))
+    }
+}
+
 /// Manifest format for api packages
 #[derive(Debug, Clone, Hash, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
 pub struct PackageManifest {
     /// Type of the package
     pub r#type: PackageType,
```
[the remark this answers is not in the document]
asmello: @mara-schulke I don't have context on this one, any reason not to change this?

src/command.rs
```diff
@@ -0,0 +1,304 @@
+// Copyright 2023 Helsing GmbH
```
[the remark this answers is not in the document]
asmello: Now I've done all the hard work of reconciling with main... 😅 

More seriously though, the reason I refactored this is because `lib.rs` and `main.rs` have exact duplicates of the `cmd::install()` code, and now that this has grown more complex it's super iffy to keep the implementations separate. I _could_ fix this in a later PR (and if I'd known how much this would've conflicted with Kim's changes I'd have done so), but is it really worth reverting at this point? (maybe it is, but don't want to do this unless you feel strongly)

docs/src/faq.md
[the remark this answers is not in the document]
mara-schulke: It is to have a strict structure for declaring manifests (which always consists of [package] + [dependencies]) – i thought this would a) simplify the learning curve as someone doesn't need to know when which section is mandatory and when not.

This creates symmetry between libraries, apis and implementations (+ im somewhat confident that we could enable lib to lib deps in some distant future)

src/command.rs
```diff
@@ -0,0 +1,304 @@
...
+    manifest::{Dependency, Manifest, PackageManifest},
+    package::{DependencyGraph, PackageName, PackageStore, PackageType},
+    registry::{Artifactory, Registry, RegistryUri},
+    Language,
+};
+
+use std::{env, path::Path, sync::Arc};
+
+use eyre::{ensure, Context};
+use semver::{Version, VersionReq};
```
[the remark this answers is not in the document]
asmello: Do you mean having the std and 3rd party imports separate? I can change that.

src/command.rs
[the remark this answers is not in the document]
mara-schulke: Agreed, the diff is not really understandable. One reason for doing this is that `buffrs::build` reuses the install command now i think

src/command.rs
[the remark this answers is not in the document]
mara-schulke: grouping in terms of using `{..}` or creating groups through newlines?

src/command.rs
```diff
@@ -0,0 +1,304 @@
...
+/// Installs dependencies
+pub async fn install(credentials: Credentials) -> eyre::Result<()> {
+    let credentials = Arc::new(credentials);
+
+    let manifest = Manifest::read().await?;
+
+    let lockfile = Lockfile::read_or_default().await?;
+
+    let dependency_tree =
+        DependencyGraph::from_manifest(&manifest, &lockfile, &credentials).await?;
```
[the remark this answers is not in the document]
mara-schulke: this is already planned :) i spoke with @Tpt a while ago about it

src/command.rs
```diff
@@ -0,0 +1,304 @@
...
+
+    let mut locked = Vec::new();
+
+    #[async_recursion]
+    async fn process_recursively(
```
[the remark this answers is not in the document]
mara-schulke: agree, feels somewhat hard to track

src/lib.rs
```diff
@@ -1,29 +1,40 @@
...
 #![doc = include_str!("../README.md")]
 
+/// CLI command implementations
+pub mod command;
 /// Credential management
 pub mod credentials;
 /// Code generator
 #[cfg(feature = "build")]
 pub mod generator;
+/// Lockfile Implementation
```
mara-schulke, line 24:
```suggestion
/// Lockfile implementation
```

src/lock.rs
asmello: In practice all we care about is the data since we only use one algorithm (and even if we introduce another one in the future, I don't think we'd ever mix algorithm types together). If we ever introduce other supported algorithms we can update this to have some sort of artificial ordering between algorithm types, just to make it extra robust, but for now it's literally pointless as we'd be comparing Self::SHA256 with itself.

src/lock.rs
asmello: fair enough!

src/lock.rs
mara-schulke: A `Dependency` holds a version requirement which can also be a range – im pretty sure this works in the future

mara-schulke: The point is that the `Dependency` type is used to fetch the dependency from the registry – if we internally lock the version -> map it as a pinned version im pretty sure it is ok, as this is _the_ job of the lockfile

src/manifest.rs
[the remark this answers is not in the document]
mara-schulke: no feel free

src/package.rs
```diff
@@ -352,8 +340,24 @@ impl TryFrom<Bytes> for Package {
 pub enum PackageType {
     /// A library package containing primitive type definitions
     Lib,
-    /// A api package containing message and service definition
+    /// An api package containing message and service definition
     Api,
+    /// An implementation package that implements an api or library
+    ///
+    /// Note: Implementation packages can't be published via Buffrs
+    Impl,
```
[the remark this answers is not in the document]
mara-schulke: sorta, i introduced it to create a symmetry which makes the impl internally way easier as you otherwise have to check "package is none" instead of "package is an impl"

src/lock.rs
asmello: I don't think so, it's always valid to produce a `Dependency` from a` LockedPackage`, since a `LockedPackage` will always have a pinned version. It's the reverse path that isn't valid.

(that said I don't like the API we use for `Registry::download()` expecting a `Dependency`, feels like it should be some more general representation of a package locator — but that's a problem for another PR).

src/lock.rs
asmello: Hmm if we're writing our own deserializer, is there a point to having `RawLockfile` at all? It's only there as an intermediate format that can be directly serialized/deserialized with the default toml serde implementation.

I think I'll keep it as is for now since this is a private struct and I don't expect anything will break _outside this module_ if we change how ser/de works later, but correct me if I'm wrong.

src/command.rs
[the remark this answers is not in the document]
asmello: I assumed the latter since the bracket grouping is how `rustfmt` likes things.

src/command.rs
```diff
@@ -0,0 +1,304 @@
...
+        Some(name) => name,
+        None => std::env::current_dir()?
+            .file_name()
+            .wrap_err(DIR_ERR)?
+            .to_str()
+            .wrap_err(DIR_ERR)?
+            .parse()?,
+    };
+
+    const VERSION: Version = Version::new(0, 1, 0);
```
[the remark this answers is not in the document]
asmello: I'm trying to avoid changing things about pre-existing code, but this is tiny enough I'll just do it.

src/command.rs
```diff
@@ -0,0 +1,304 @@
...
+        dependencies: vec![],
+    };
+
+    manifest.write().await?;
+
+    PackageStore::create().await
+}
+
+/// Adds a dependency to this project
+pub async fn add(registry: RegistryUri, dependency: String) -> eyre::Result<()> {
```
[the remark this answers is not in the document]
asmello: Ok, will also pick up this one.

src/command.rs
```diff
@@ -0,0 +1,304 @@
...
+    let credentials = Arc::new(credentials);
+
+    let manifest = Manifest::read().await?;
+
+    let lockfile = Lockfile::read_or_default().await?;
+
+    let dependency_tree =
+        DependencyGraph::from_manifest(&manifest, &lockfile, &credentials).await?;
+
+    let mut locked = Vec::new();
```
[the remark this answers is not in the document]
asmello: Probably! This whole algorithm will need changes, but I tried to make the general structure future-aware at least.

src/command.rs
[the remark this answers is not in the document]
asmello: This isn't a resolver, it's just traversing the dependency graph that was build previously. I can improve the naming but I don't think it's a good idea to unify the two or we'd end up with a situation where we fail validation but have already installed some packages.

src/lock.rs
asmello: Oh good catch, this is leftover from a previous version of the Digest representation.

src/command.rs
```diff
@@ -0,0 +1,304 @@
...
+pub async fn init(r#type: PackageType, name: Option<PackageName>) -> eyre::Result<()> {
+    ensure!(
+        !Manifest::exists().await?,
+        "Cant initialize existing project"
+    );
+
+    const DIR_ERR: &str = "Failed to read current directory name";
+
+    let name = match name {
+        Some(name) => name,
```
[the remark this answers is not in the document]
asmello: I think Jon was proposing
```rust
    let name = if let Some(name) = name {
        name
    } else {
        std::env::current_dir()?
            .file_name()
            .wrap_err(DIR_ERR)?
            .to_str()
            .wrap_err(DIR_ERR)?
            .parse()?
    };
```

Which is still a bit awkward but slightly better.

Could also use
```rust
    fn curr_dir_name() -> eyre::Result<PackageName> {
        std::env::current_dir()?
            .file_name()
            .wrap_err(DIR_ERR)?
            .to_str()
            .wrap_err(DIR_ERR)?
            .parse()
    }

    let name = name.map(Result::Ok).unwrap_or_else(curr_dir_name)?;
```

But I'm not sure that's an improvement.

src/command.rs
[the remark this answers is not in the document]
asmello: I also considered implementing a `DependencyGraph::traverse()` method and providing it a closure here but couldn't figure out a clean way to keep track of the prefix with that approach.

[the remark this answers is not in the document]
asmello: TBH I kinda want to reimplement this with an iterative approach to avoid the dependency on `async_recursion` but I figured this would be easier to reason about for a first implementation.

src/command.rs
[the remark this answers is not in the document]
mara-schulke: How about unwrap_or_else?

docs/src/faq.md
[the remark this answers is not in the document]
jonhoo: But Cargo _doesn't_ require `[dependencies]` if it's empty.

I guess what I'm trying to understand is the value of always including `[dependencies]` — it doesn't feel like something that aids learning, because now you have to put a thing in a file even if you don't need said thing.

Now, it'd be different if you asserted that every project will have at least one dependency, but I don't think that's true?

src/command.rs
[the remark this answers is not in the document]
jonhoo: The latter — splitting `use`s into newline-separated parts depending on `crate` vs `std` vs everything else.

src/manifest.rs
```diff
@@ -21,9 +33,9 @@ pub const MANIFEST_FILE: &str = "Proto.toml";
 /// This contains the exact structure of the `Proto.toml` and skips
 /// empty fields.
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
-pub struct RawManifest {
-    pub package: Option<PackageManifest>,
-    pub dependencies: Option<DependencyMap>,
+struct RawManifest {
+    package: PackageManifest,
+    dependencies: DependencyMap,
```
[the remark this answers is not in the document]
jonhoo: I think there's something I'm missing here — isn't this the `Proto.toml` file? Surely there isn't value in ever requiring an empty section to have to be present in such a file when its presence doesn't make a difference? Maybe the argument is that `[package]` will _never_ be empty, and then I totally agree it shouldn't be `#[serde(default)]` to enforce its presence, but `[dependencies]` surely can be empty (eg, for root protos).

src/lock.rs
jonhoo: My concern is more-so that if you _do_ introduce other digest algorithms, you almost certainly won't then remember to update this `impl`. So why not just have it take into account digest from the beginning (same way `Eq` does)?

docs/src/faq.md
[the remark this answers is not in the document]
mara-schulke: Okay agreed, making the package section mandatory was the important change - the dependency section can be sensibly defaulted to an empty list

src/command.rs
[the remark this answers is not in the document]
asmello: Fair enough. I'll avoid this in the future. For what it is worth the only function I touched here is `install()`. Well at least aside from your nits I actioned in the others.

src/manifest.rs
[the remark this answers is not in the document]
mara-schulke: Yes agreed!

src/lock.rs
asmello: Ok, after some sleep I figured I can implement a trivial cmp for Algorithm and that will leave this implementation here stable against future additions. Plus, the other cmp implementation will be sensitive to new additions so no one will forget to update it.

docs/src/faq.md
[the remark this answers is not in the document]
asmello: Made the change.

asmello:
Ok I believe I've accounted for all feedback now!

src/manifest.rs
```diff
@@ -21,9 +33,10 @@ pub const MANIFEST_FILE: &str = "Proto.toml";
 /// This contains the exact structure of the `Proto.toml` and skips
 /// empty fields.
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
-pub struct RawManifest {
-    pub package: Option<PackageManifest>,
-    pub dependencies: Option<DependencyMap>,
+struct RawManifest {
+    package: PackageManifest,
+    #[serde(default)]
+    dependencies: DependencyMap,
```
mara-schulke, line 39: is this now actually optional?

mara-schulke (commented):
Is the dependency section now optional?

asmello:
> Is the dependency section now optional?

Yes it is!

PR #48, opened 2023-09-13
Add package subcommand

### LINKED ISSUE #47: Add package subcommand
Currently the only way to produce a package is with the `publish` subcommand. This relies on access to a functioning Artifactory instance to work.

In some cases it may be useful to just produce the tgz and distribute it some other way.

asmello:
Fixes #47


src/main.rs
```diff
@@ -41,6 +41,18 @@ enum Command {
...
+    /// Packages this api
+    #[clap(alias = "pack")]
+    Package {
+        /// Destination directory for the release
+        #[clap(long)]
+        #[arg(default_value = ".")]
+        directory: String,
+        /// Generate package but do not write it to filesystem
+        #[clap(long)]
+        dry_run: bool,
```
mara-schulke, line 53: I would say this is not needed – people can just delete the tgz

src/main.rs
```diff
@@ -41,6 +41,18 @@ enum Command {
         package: PackageId,
     },
 
+    /// Packages this api
+    #[clap(alias = "pack")]
+    Package {
+        /// Destination directory for the release
+        #[clap(long)]
+        #[arg(default_value = ".")]
+        directory: String,
```
mara-schulke, line 50: Can we name it `--output-directory`? Directory could refer to the package directory as well / could be misleading

src/main.rs
```diff
@@ -41,6 +41,18 @@ enum Command {
         package: PackageId,
     },
 
+    /// Packages this api
```
mara-schulke, line 44:
```suggestion
    /// Exports the current package into a distributable tgz archive
```

src/main.rs
```diff
@@ -234,6 +247,23 @@ mod cmd {
         manifest.write().await
     }
 
+    pub async fn package(directory: String, dry_run: bool) -> eyre::Result<()> {
```
mara-schulke, line 250: This one is missing a doc comment

mara-schulke (approved):

src/main.rs
asmello: Makes sense, will update

src/main.rs
asmello: I think this is nice to have if you just want to validate the manifest and don't want to risk overwriting something. Unless you feel it's not worth the complexity I'd prefer to keep it.

PR #49, opened 2023-09-13
Preserve subfolders in generated packages

### LINKED ISSUE #39: Preserve subfolders in packages
Currently when creating a buffrs package with in this buffrs project:

```
Proto.toml
proto/
  top_level.proto
  nested/
    nested.proto
```

the tgz structure looks like this:

```
Proto.toml
top_level.proto
nested.proto
```

when it should look like this:

```
Proto.toml
top_level.proto
nested/
  nested.proto
```

asmello:
Fixes #39

### DIFF
--- a/tests/lib.rs
+++ b/tests/lib.rs
@@ -1,11 +1,12 @@
 use std::{
     fs,
+    io::ErrorKind,
     path::{Path, PathBuf},
 };
 
 use assert_fs::TempDir;
 use fs_extra::dir::{get_dir_content, CopyOptions};
-use pretty_assertions::assert_eq;
+use pretty_assertions::{assert_eq, assert_str_eq};
 
 mod cmd;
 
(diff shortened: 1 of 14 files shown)


tests/lib.rs
```diff
@@ -133,10 +134,15 @@ impl VirtualFileSystem {
...
-                &fs::read(&expected).expect("file not found"),
-                &fs::read(&actual).expect("file not found"),
-            );
+            match fs::read_to_string(&expected) {
+                Ok(expected_text) => assert_str_eq!(
+                    expected_text,
+                    fs::read_to_string(&actual).expect("file not found")
+                ),
+                // TODO(amello): make tgz deterministic so we can validate binary files
+                Err(err) if matches!(err.kind(), ErrorKind::InvalidData) => (),
```
mara-schulke, line 143: Can we filter tar / tgz files from the list of files to compare instead of catching this error?

mara-schulke:
Two general comments:

1. Is there a reason why there is a diff in the tests for add/remove?
2. `proto/vendor` is the vendored package store -> `proto/vendor/*.proto` is illegal / out of place as it would be packaged up and installed into this file if it came from the `external` package: `proto/vendor/external/external.proto`

mara-schulke:
One last thing: can we remove the symlinks? I thought they were a good idea initially but it turns out to be really confusing 🙈

mara-schulke (approved):

PR #51, opened 2023-09-14
Make packages deterministic

### LINKED ISSUE #50: Make packages deterministic
Currently it appears the generated tar packages are sensitive to filesystem metadata (particularly last access timestamps etc.), which means two packages built in two different computers might have a different checksum even if the actual contents are identical.

In order to make building and publishing a reproducible process (which should pave the way for better testing and supply chain security), generating the tar packages should be a deterministic process which depends only on the protobuf definitions.

asmello:
Fixes #50


tests/lib.rs
```diff
@@ -133,16 +135,31 @@ impl VirtualFileSystem {
 
             println!("\n-- {} –-\n", file.display());
 
-            if let Some(extension) = file.extension() {
-                if extension == "tgz" {
-                    continue; // TODO(amello): make packaging deterministic so we can validate the files
+            match fs::read_to_string(&expected) {
```
mara-schulke, line 138: can we do the comparison by file ext instead of using the errorkind?

so having two branches depending on the extension (text {*}, bin {tgz}) instead of trying one and then the other – that would also be more explicit on what / why stuff happens

eg having a `enum FileType` with a `from_extension` or similar (not worring to much about the naming but rather encoding what happens into the typesystem

tests/lib.rs
```diff
@@ -133,16 +135,31 @@ impl VirtualFileSystem {
...
+                Ok(expected_text) => {
+                    let actual_text = fs::read_to_string(&actual).expect("file not found");
+                    assert_str_eq!(expected_text, actual_text);
                 }
+                Err(err) if matches!(err.kind(), ErrorKind::InvalidData) => {
+                    // binary file, use checksum comparison
+
+                    fn hash_file(path: impl AsRef<Path>) -> String {
```
mara-schulke, line 146: can we use ring for the hash? there is a feature branch that also relies on ring and it would be nice to dedup dependencies

Cargo.toml
```diff
@@ -50,6 +50,7 @@ tracing = "0.1"
 tracing-subscriber = "0.3"
 url = { version = "2.4", features = ["serde"] }
 walkdir = "2"
+sha2 = "0.10.7"
```
mara-schulke, line 53: sha _would_ be a dev dependency btw – but #38 relies on ring so if you use ring we could keep it a direct dep

tests/lib.rs
asmello: Missed that dependency, will switch over.

tests/lib.rs
asmello: I wanted to avoid having to know about all file types ahead of time, in order to make this utility more flexible. That said, on a second thought, one issue with checking UTF8 encoding is the possibility of running into a binary file that happens to be UTF8 compatible by sheer chance.

Since we control the test files I think it's ok to go the enumeration route. Will make the change.

mara-schulke (approved):

PR #52, opened 2023-09-14
Improve artifactory response validations

### LINKED ISSUE #40: Decompression fails with invalid auth config
```
$ buffrs install
Error:
   0: Failed to install dependencies
   1: Failed to decompress package
   2: invalid gzip header

Location:
   /Users/mara.schulke/.cargo/registry/src/index.crates.io-6f17d22bba15001f/buffrs-0.5.0/src/package.rs:300

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
```

This occurs when the registry configuration is nonesense:

```
───────┼────────────────────────────────────────────────────
       │ File: $HOME/.buffrs/credentials.toml
───────┼────────────────────────────────────────────────────
   1   │ [artifactory]
   2   │ url = "https://org.jfrog.io/artifactory"
   3   │ username = "foo"
   4   │ password = "some-token"
```

The request does not result in a 4xx or 5xx code as the url is redirected to https://landing.jfrog.com/reactivate-server/org

---

The error is very cryptic and we should verify in the `artifactory.rs` the the registry did return the correct mime type.

asmello:
Fixes #40


mara-schulke:
This would need to check mime types aswell!! Otherwise the weird "invalid gzip" error wont go away, if anything else than binary is returned

asmello:
Ack, for the common case described in #40 preventing a redirect is sufficient, but I added an extra check for the mime type as well in the download method implementation.

mara-schulke (approved):

PR #58, opened 2023-09-18
Decouple registry configuration from credentials / enable multiple registry support

kihehs:
This enables logging into and downloading protobufs from more than one registry.

### DIFF
--- a/src/credentials.rs
+++ b/src/credentials.rs
@@ -2,10 +2,15 @@
 
 use eyre::{Context, ContextCompat};
 use serde::{Deserialize, Serialize};
-use std::{env, path::PathBuf};
+use std::{
+    collections::HashMap,
+    env,
+    ops::{Deref, DerefMut},
+    path::PathBuf,
+};
 use tokio::fs;
 
-use crate::registry::ArtifactoryConfig;
+use crate::manifest::{RegistryToken, RegistryUri};
 
 /// Global configuration directory for `buffrs`
 pub const BUFFRS_HOME: &str = ".buffrs";
@@ -13,10 +18,23 @@ pub const BUFFRS_HOME: &str = ".buffrs";
 pub const CREDENTIALS_FILE: &str = "credentials.toml";
 
 /// Credential store for storing authentication data
-#[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[derive(Debug, Default, Clone)]
 pub struct Credentials {
-    /// Artifactory credentials
-    pub artifactory: Option<ArtifactoryConfig>,
+    credentials: HashMap<RegistryUri, RegistryToken>,
+}
+
+impl Deref for Credentials {
+    type Target = HashMap<RegistryUri, RegistryToken>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.credentials
+    }
+}
+
+impl DerefMut for Credentials {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.credentials
+    }
 }
 
 impl Credentials {
@@ -41,7 +59,10 @@ impl Credentials {
             .await
             .wrap_err("Failed to read credentials")?;
 
-        toml::from_str(&toml).wrap_err("Failed to parse credentials")
+        let raw: RawCredentialCollection =
+            toml::from_str(&toml).wrap_err("Failed to parse credentials")?;
+
+        Ok(raw.into())
     }
 
     /// Writes the credentials to the file system
@@ -54,7 +75,9 @@ impl Credentials {
         .await
         .ok();
 
-        fs::write(Self::location()?, toml::to_string(&self)?.into_bytes())
+        let data: RawCredentialCollection = self.clone().into();
+
+        fs::write(Self::location()?, toml::to_string(&data)?.into_bytes())
             .await
             .wrap_err("Failed to write credentials")
     }
@@ -70,5 +93,41 @@ impl Credentials {
         };
 
         Ok(credentials)
+    }
+}
+
+/// Credential store for storing authentication data. Serialization type.
+#[derive(Serialize, Deserialize)]
+pub struct RawCredentialCollection {
+    credentials: Vec<RawRegistryCredentials>,
+}
+
+/// Credentials for a single registry. Serialization type.
+#[derive(Serialize, Deserialize)]
+struct RawRegistryCredentials {
+    uri: RegistryUri,
+    token: RegistryToken,
+}
+
+impl From<RawCredentialCollection> for Credentials {
+    fn from(value: RawCredentialCollection) -> Self {
+        let credentials = value
+            .credentials
+            .into_iter()
+            .map(|it| (it.uri, it.token))
+            .collect();
+        Self { credentials }
+    }
+}
+
+impl From<Credentials> for RawCredentialCollection {
+    fn from(value: Credentials) -> Self {
+        let credentials = value
+            .credentials
+            .into_iter()
+            .map(|(uri, token)| RawRegistryCredentials { uri, token })
+            .collect();
+
+        Self { credentials }
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -34,7 +34,7 @@ pub fn build(language: Language) -> eyre::Result<()> {
     async fn install() -> eyre::Result<()> {
         let credentials = Credentials::read().await?;
 
-        let artifactory = Artifactory::from(
+        let artifactory = Artifactory::new(
             credentials
                 .artifactory
                 .wrap_err("Artifactory configuration is required")?,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -42,8 +42,6 @@ async fn install() -> eyre::Result<()> {
     let credentials = Credentials::read().await?;
     let manifest = Manifest::read().await?;
 
-    let artifactory = Artifactory::new(credentials);
-
     let mut install = Vec::new();
 
     for dependency in manifest.dependencies {
@@ -60,7 +58,13 @@ async fn install() -> eyre::Result<()> {
             }
         }
 
-        install.push(PackageStore::install(dependency, artifactory.clone()));
+        let artifactory =
+            Artifactory::new(credentials.clone(), dependency.manifest.registry.clone());
+
+        install.push(PackageStore::install(
+            dependency.clone(),
+            artifactory.clone(),
+        ));
     }
 
     futures::future::try_join_all(install)
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,61 +17,55 @@ pub mod registry;
 #[cfg(feature = "build")]
 pub use generator::Language;
 
+use credentials::Credentials;
+use eyre::{ContextCompat, WrapErr};
+use manifest::Manifest;
+use package::PackageStore;
+use registry::Artifactory;
+
 /// Cargo build integration for buffrs
 ///
 /// Important: Only use this inside of cargo build scripts!
 #[cfg(feature = "build")]
 pub fn build(language: Language) -> eyre::Result<()> {
-    use credentials::Credentials;
-    use eyre::ContextCompat;
-    use eyre::WrapErr;
-    use manifest::Manifest;
-    use package::PackageStore;
-    use registry::Artifactory;
-
     println!("cargo:rerun-if-changed={}", PackageStore::PROTO_VENDOR_PATH);
-
-    async fn install() -> eyre::Result<()> {
-        let credentials = Credentials::read().await?;
-
-        let artifactory = Artifactory::new(
-            credentials
-                .artifactory
-                .wrap_err("Artifactory configuration is required")?,
-        );
-
-        let manifest = Manifest::read().await?;
-
-        let mut install = Vec::new();
-
-        for dependency in manifest.dependencies {
-            if let Ok(pkg) = PackageStore::resolve(&dependency.package).await {
-                let pkg = pkg.package.wrap_err_with(|| {
-                    format!(
-                        "required package entry in manifest of {} to be present",
-                        dependency.package
-                    )
-                })?;
-
-                if dependency.manifest.version.matches(&pkg.version) {
-                    continue;
-                }
-            }
-
-            install.push(PackageStore::install(dependency, artifactory.clone()));
-        }
-
-        futures::future::try_join_all(install)
-            .await
-            .wrap_err("Failed to install missing dependencies")?;
-
-        Ok(())
-    }
 
     let rt = tokio::runtime::Runtime::new()?;
 
     rt.block_on(install())?;
     rt.block_on(generator::generate(language))?;
+
+    Ok(())
+}
+
+async fn install() -> eyre::Result<()> {
+    let credentials = Credentials::read().await?;
+    let manifest = Manifest::read().await?;
+
+    let artifactory = Artifactory::new(manifest.clone(), credentials);
+
+    let mut install = Vec::new();
+
+    for dependency in manifest.dependencies {
+        if let Ok(pkg) = PackageStore::resolve(&dependency.package).await {
+            let pkg = pkg.package.wrap_err_with(|| {
+                format!(
+                    "required package entry in manifest of {} to be present",
+                    dependency.package
+                )
+            })?;
+
+            if dependency.manifest.version.matches(&pkg.version) {
+                continue;
+            }
+        }
+
+        install.push(PackageStore::install(dependency, artifactory.clone()));
+    }
+
+    futures::future::try_join_all(install)
+        .await
+        .wrap_err("Failed to install missing dependencies")?;
 
     Ok(())
 }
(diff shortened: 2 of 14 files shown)


src/manifest.rs
```diff
@@ -144,5 +156,119 @@ pub struct DependencyManifest {
...
+        write!(f, "{}", self.0)
+    }
+}
+
+impl FromStr for RegistryUri {
+    type Err = eyre::Error;
+
+    fn from_str(value: &str) -> Result<Self, Self::Err> {
+        let slf = Self(Url::from_str(value)?);
+        slf.sanity_check_url()?;
```
mara-schulke, line 192: This is dangerous as it assumes *every* registry ends with `/artifactory` – could you please move it back to the artifactory code + make it only happening if the host is `*.jfrog.io`?

src/manifest.rs
```diff
@@ -144,5 +156,119 @@ pub struct DependencyManifest {
...
+            self.0.path().ends_with("/artifactory"),
+            "The url must end with '/artifactory'"
+        );
+
+        Ok(())
+    }
+}
+
+#[derive(Debug, Default, Clone, Hash, Serialize, Deserialize, PartialEq, Eq)]
+pub struct RegistryToken(pub String);
```
mara-schulke, line 222: Adds no value as we cant assume structure on tokens

src/manifest.rs
```diff
@@ -144,5 +156,119 @@ pub struct DependencyManifest {
...
+impl FromStr for RegistryToken {
+    type Err = eyre::Error;
+
+    fn from_str(value: &str) -> Result<Self, Self::Err> {
+        Ok(Self(value.to_owned()))
+    }
+}
+
+#[derive(Debug, Clone, Hash, Serialize, Deserialize, PartialEq, Eq)]
+pub struct Repository(pub String);
```
mara-schulke, line 247: Same here

---

A good indicator for new types imo to introduce them once you have logic (such as parsing, validating etc, tied to a primitive type – `RegistryToken` and `Repository` dont contain any logic + create a layer of indirection

src/manifest.rs
```diff
@@ -144,5 +156,119 @@ pub struct DependencyManifest {
...
+}
+
+impl From<&str> for Repository {
+    fn from(value: &str) -> Self {
+        Self(value.into())
+    }
+}
+
+/// Authentication data and settings for an artifactory registry
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
```
mara-schulke, line 270:
```suggestion
#[derive(Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
```

credentials should never be printed – if you want them to be printed you should impl debug manually / omitting the token

src/manifest.rs
```diff
@@ -144,5 +156,119 @@ pub struct DependencyManifest {
...
+pub struct RawRegistryCredentials {
```
mara-schulke, line 271: Valid type, but belongs in the `credential.rs`

src/registry/artifactory.rs
```diff
@@ -1,36 +1,50 @@
...
     /// Pings artifactory to ensure registry access is working
     pub async fn ping(&self) -> eyre::Result<()> {
-        let repositories_uri: Url = {
-            let mut url = self.0.url.clone();
-            url.set_path(&format!("{}/api/repositories", url.path()));
-            url
+        let token = self
+            .credentials
+            .get(&self.registry)
+            .wrap_err("no token found for the given registry")?;
```
mara-schulke, line 33: This is not an error – auth is optional!

src/registry/artifactory.rs
```diff
@@ -1,36 +1,50 @@
...
             .redirect(reqwest::redirect::Policy::none())
             .build()
             .wrap_err("client error")?
-            .get(repositories_uri.clone())
-            .header(
-                "X-JFrog-Art-Api",
-                self.0.password.clone().unwrap_or_default(),
-            )
+            .get(repositories_uri.as_str())
+            .header("X-JFrog-Art-Api", &**token)
```
mara-schulke, line 47: this call is optional

nit: can we make the header name a constant?

src/registry/artifactory.rs
```diff
@@ -76,12 +90,13 @@ impl Registry for Artifactory {
             }
         );
 
-        let artifact_uri: Url = {
-            let mut url = self.0.url.clone();
+        let artifact_uri = {
+            let path = dependency.manifest.registry.path().to_owned();
```
mara-schulke, line 94: Ah i see why you give it the whole credentials struct. Could we still wrap it in an arc?

src/registry/artifactory.rs
```diff
@@ -91,17 +106,29 @@ impl Registry for Artifactory {
...
+                // Go on without auth.
+                reqwest::Client::builder()
+                    .redirect(reqwest::redirect::Policy::none())
+                    .build()
+                    .wrap_err("client error")?
+                    .get((*artifact_uri).clone())
+                    .send()
+                    .await?
+            }
+        };
```
mara-schulke, line 131: you could dedup code by making the builder mutable and only appending the header if you happen to have a token

src/registry/artifactory.rs
```diff
@@ -132,10 +159,10 @@ impl Registry for Artifactory {
     }
 
     /// Publishes a package to artifactory
-    async fn publish(&self, package: Package, repository: String) -> eyre::Result<()> {
+    async fn publish(&self, package: Package, repository: Repository) -> eyre::Result<()> {
         let artifact_uri: Url = format!(
             "{}/{}/{}/{}-{}.tgz",
-            self.0.url,
+            self.registry,
```
mara-schulke, line 165: Indeed this should be an argument, sorry for telling you otherwise

src/registry/artifactory.rs
```diff
@@ -144,19 +171,33 @@ impl Registry for Artifactory {
...
+                    .redirect(reqwest::redirect::Policy::none())
+                    .build()
+                    .wrap_err("client error")?
+                    .put(artifact_uri.clone())
+                    .body(package.tgz)
+                    .send()
+                    .await
+                    .wrap_err("Failed to upload release to artifactory")?
+            }
+        };
```
mara-schulke, line 200: Same dedup here

src/registry/local.rs
```diff
@@ -137,10 +140,14 @@ mod tests {
             package_bytes
         );
 
+        let registry_uri = RegistryUri::from_str("http://some-registry/artifactory")
+            .expect("Failed to parse registry URL");
```
mara-schulke, line 144: Im the registry uri type is fine – if it does validate the `*.jfrog.io` rule + protocol assumptions while parsing!

tests/cmd/login/mod.rs
```diff
@@ -6,7 +6,7 @@ fn fixture() {
 
     crate::cli!()
         .arg("login")
-        .arg("--url")
+        .arg("--registry")
```
mara-schulke, line 9: This will be a breaking change for the kraken integration -> requires a mr in kraken

tests/cmd/logout/out/$HOME/.buffrs/credentials.toml
```diff
@@ -0,0 +1 @@
+credentials = []
```
mara-schulke, line 1: I think this should be completely empty in the best case, i think serde can to that with `skip`

src/main.rs
```diff
@@ -367,14 +367,12 @@ mod cmd {
                 .read_line(&mut raw)
                 .wrap_err("Failed to read token")?;
 
-            raw = raw.trim().to_owned();
-
-            raw
+            RegistryToken::from_str(raw.trim())?
```
mara-schulke, line 370: I (personally) find `.parse()` more idiomatic – often you can get around the turbofish syntax if you parse a value thats an arg to something thats typed.

So here `Artifactory::new` is typed so you could get around explicitly calling from_str

src/credentials.rs
```diff
@@ -2,21 +2,39 @@
...
 /// Credential store for storing authentication data
-#[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[derive(Debug, Default, Clone)]
 pub struct Credentials {
-    /// Artifactory credentials
-    pub artifactory: Option<ArtifactoryConfig>,
+    credentials: HashMap<RegistryUri, RegistryToken>,
+}
+
+impl Deref for Credentials {
```
mara-schulke, line 26: Overusing deref / implementing inhertiance through deref is an antipattern – see https://rust-unofficial.github.io/patterns/anti_patterns/deref.html

src/credentials.rs
```diff
@@ -2,21 +2,39 @@
...
 /// Filename of the credential store
 pub const CREDENTIALS_FILE: &str = "credentials.toml";
 
 /// Credential store for storing authentication data
-#[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[derive(Debug, Default, Clone)]
 pub struct Credentials {
-    /// Artifactory credentials
-    pub artifactory: Option<ArtifactoryConfig>,
+    credentials: HashMap<RegistryUri, RegistryToken>,
```
mara-schulke, line 23: this can just be pub

mara-schulke (changes requested):
Left a couple of comments 😊 
Lmk if you have questions!

mara-schulke:
Note that your sanity check is dangerous atm as `/artifactory` as a suffix is only true for `*.jforg.io` hosts!

tests/cmd/logout/out/$HOME/.buffrs/credentials.toml
mara-schulke: Also a nit (good become a "good first issue")

src/manifest.rs
kihehs: Removed the newtype.

src/manifest.rs
kihehs: Moved to credentials.rs.

src/registry/artifactory.rs
```diff
@@ -1,36 +1,50 @@
...
+    credentials::Credentials,
+    manifest::{Dependency, RegistryUri, Repository},
+    package::Package,
+};
 
 /// The registry implementation for artifactory
 #[derive(Debug, Clone)]
-pub struct Artifactory(Arc<ArtifactoryConfig>);
+pub struct Artifactory {
+    credentials: Credentials,
```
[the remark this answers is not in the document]
kihehs: Refactored to use `Arc<Credentials>` instead.

src/registry/artifactory.rs
kihehs: Good catch! Fixed!

src/registry/artifactory.rs
kihehs: I made the header a constant, which it definitely should be. I also made it so that the header is only set when the token exists.

src/registry/artifactory.rs
kihehs: Agreed, that's on the horizon. I just copied for now because it allowed me to move faster.

src/registry/local.rs
kihehs: Yup, I'll give it a constructor instead.

tests/cmd/login/mod.rs
kihehs: Good point. Should I change it back? I thought it would be nice to have everything named `--registry`, as it's a more descriptive name, but it may not be a good idea.

tests/cmd/logout/out/$HOME/.buffrs/credentials.toml
kihehs: It's easy to fix. Do you think it should be left as a good first issue, or do I just fix it? I don't mind either way.

src/main.rs
kihehs: I already removed the whole type, so this is moot :)

src/credentials.rs
kihehs: Good point. I'll change this.

src/registry/artifactory.rs
kihehs: The header method consumes self, so it's not quite as pretty as declaring the builder mut, but it's okay..

```rust
pub fn header<K, V>(self, key: K, value: V) -> RequestBuilder
```

I can still do this:

```rust
        let response = {
            let builder = reqwest::Client::builder()
                .redirect(reqwest::redirect::Policy::none())
                .build()
                .wrap_err("client error")?
                .get(repositories_uri.as_str());

            let builder = if let Some(token) = self.credentials.credentials.get(&self.registry) {
                builder.header(JFROG_AUTH_HEADER, token)
            } else {
                builder
            };

            builder.send().await?
        };
```

mara-schulke: I think that is much more readable in terms of "what is the code doing"

tests/cmd/login/mod.rs
mara-schulke: No its completly valid! Just a thing foir the FAQ + to keep in mind that kraken has to be adopted

tests/cmd/logout/out/$HOME/.buffrs/credentials.toml
mara-schulke: Up to u :)

docs/src/faq.md
```diff
@@ -0,0 +1,32 @@
+# FAQ
+
+
+## Breaking Changes
```
mara-schulke, line 4: It would be cool to not write the changelog here but rather "why do i have error x and how to i fix it".

So one thing could be "my invocation of buffrs login is broken" and the answer would be that we renamed `--url` to `--registry`

tests/cmd/login/mod.rs
kihehs: Yep, I will update kraken soon.

src/credentials.rs
```diff
@@ -2,21 +2,20 @@
 
 use eyre::{Context, ContextCompat};
 use serde::{Deserialize, Serialize};
-use std::{env, path::PathBuf};
+use std::{collections::HashMap, env, path::PathBuf};
 use tokio::fs;
 
-use crate::registry::ArtifactoryConfig;
+use crate::manifest::RegistryUri;
```
mara-schulke, line 8: maybe this types is best located in the registry `mod.rs`? it neither strictly belongs in the manifest nor the credentials mod

src/credentials.rs
```diff
@@ -2,21 +2,20 @@
...
 /// Filename of the credential store
 pub const CREDENTIALS_FILE: &str = "credentials.toml";
 
 /// Credential store for storing authentication data
-#[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[derive(Debug, Default, Clone)]
 pub struct Credentials {
-    /// Artifactory credentials
-    pub artifactory: Option<ArtifactoryConfig>,
+    pub registry_tokens: HashMap<RegistryUri, String>,
```
mara-schulke, line 18: This is missing a doc comment

src/credentials.rs
```diff
@@ -72,3 +76,41 @@ impl Credentials {
         Ok(credentials)
     }
 }
+
+/// Credential store for storing authentication data. Serialization type.
+#[derive(Serialize, Deserialize)]
+pub struct RawCredentialCollection {
```
mara-schulke, line 82: Is this meant to be pub?

src/credentials.rs
```diff
@@ -72,3 +76,41 @@ impl Credentials {
...
+}
+
+impl From<RawCredentialCollection> for Credentials {
+    fn from(value: RawCredentialCollection) -> Self {
+        let credentials = value
+            .credentials
+            .into_iter()
+            .map(|it| (it.uri, it.token))
+            .collect();
+        Self {
```
mara-schulke, line 100:
```suggestion

        Self {
```

mara-schulke: (you have a mixed newline style in `from` impls, thats why im suggesting)

src/lib.rs
```diff
@@ -14,64 +14,66 @@ pub mod package;
...
 #[cfg(feature = "build")]
 pub use generator::Language;
 
+use credentials::Credentials;
+use eyre::{ContextCompat, WrapErr};
+use manifest::Manifest;
+use package::PackageStore;
+use registry::Artifactory;
+
+pub const JFROG_AUTH_HEADER: &str = "X-JFrog-Art-Api";
```
mara-schulke, line 28: can this be local to the artifactory registry impl + private? this is not useful for anyone using the rust lib

src/credentials.rs
kihehs: I moved it.

src/credentials.rs
kihehs: Added!

src/credentials.rs
kihehs: It doesn't have to be, so I removed it.

src/lib.rs
kihehs: Yeah, good point. I moved it as close to the code that consumes it as possible.

mara-schulke (approved):
This looks great, thank you!

docs/src/faq.md
mara-schulke: Very good!

PR #74, opened 2023-09-22
Project types docs

asmello:
Page renamed from package types to bring impl projects into scope.


mara-schulke:
Imo this is still a "package" type – just that impls cant be published (for now)

We may even want to consider "publishing" impls later on to list services that implement a specific api version on the documentation page of an api package.

(This is mainly for searchability and insights into the api ecosystem, not for actual code distribution)

asmello:
That's fair, but is `package` an appropriate term for it? I find it confusing because in the case of `impl` it's not clear what constitutes the package, since no physical artefact exists to represent it. Even internally in the code we use the term `package` exclusively to refer to bundled projects, and `impl` projects don't qualify.

PR #80, opened 2023-09-25
Support custom root certificate via $SSL_CERT_FILE

asmello:
This enables using the system native certificate tooling to pick up on custom root certificates automatically.

I've also refactored the Artifactory registry implementation a bit. It's more efficient now, as it reuses a connection pool between requests, but still creates a new client per dependency when running the install command. This can be improved to one client per registry, which I'll probably implement in a dedicated PR.


tetigi (approved):

mara-schulke:
can we document this in the book (in the reference section) @asmello?

PR #81, opened 2023-09-26
Log when pinging

kihehs:
This helps us determine if the token is correct.

### DIFF
--- a/src/registry/artifactory.rs
+++ b/src/registry/artifactory.rs
@@ -77,8 +77,7 @@ impl Artifactory {
         let status = response.status();
 
         if !status.is_success() {
-            tracing::info!(response_header=?response.headers());
-            tracing::info!(response_body=?response.text().await?);
+            tracing::info!("Artifactory response header: {:?}", response);
         }
 
         ensure!(
--- a/src/registry/artifactory.rs
+++ b/src/registry/artifactory.rs
@@ -55,15 +55,34 @@ impl Artifactory {
 
             let builder = if let Some(token) = self.credentials.registry_tokens.get(&self.registry)
             {
+                let first_two: String = token.chars().take(2).collect();
+
+                tracing::info!(
+                    uri=?repositories_uri.as_str(),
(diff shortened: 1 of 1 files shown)


Tpt (approved):

src/registry/artifactory.rs
```diff
@@ -75,6 +75,11 @@ impl Artifactory {
         };
 
         let status = response.status();
+
+        if !status.is_success() {
+            tracing::info!(response_header=?response.headers());
```
Tpt, line 80: debug maybe?

alexespencer: If we use debug we may need to override RUST_LOG env var in CI - hopefully this should be possible?

ntrinquier: You might want to consider the [debug string](https://github.com/seanmonstar/reqwest/blob/3abcc7c4f537c16ad9937f8cc60fb23cb506ac85/src/async_impl/response.rs#L389-L397) too:

```suggestion
            tracing::debug!("artifactory response: {:?}", response);
```

Tpt: Yes. You have an example here: https://github.com/helsing-ai/buffrs/blob/3120b1a7bb16d8fd47cfbe98ced5dcfa9635df41/.github/workflows/ci.yml#L34

kihehs: We're trying to fix a bug that only appears in CI, and I'm not sure how to override RUST_LOG there.

kihehs: Regarding `debug!`, I don't see my debug logs when I run this. I seem to remember a lot of custom initialization being done to the:

```bash
RUST_LOG=debug cargo run -- login --registry https://my-registry...
export RUST_LOG=debug; cargo run -- login --registry https://my-registry...
```

(This doesn't change with `RUST_BACKTRACE=1`)

kihehs:
Closing because another merge breaks this PR pretty badly.

PR #82, opened 2023-09-26
Log when pinging

kihehs:
This helps us determine if the token is correct.

Please note that I cannot easily use `tracing::debug`, as it simply doesn't log anything. I believe this is due to how tracing_subscriber is initialized, but I'm not sure.

```rust
    tracing_subscriber::fmt()
        .compact()
        .without_time()
        .with_level(false)
        .with_file(false)
        .with_target(false)
        .with_line_number(false)
        .try_init()
        .unwrap();
```


alexespencer (approved):
We can always remove / change to debug after we've fixed this issue. LGTM?

Issue #76, opened 2023-09-22
Getting lockfile mismatch errors when editing the manifest


mara-schulke:
```
λ buffrs install
Error:
   0: Mismatched registry detected for dependency xyz - requested https://xyz.jforg.io/artifactory but lockfile requires https://xyz.jfrog.io/artifactory

Location:
   src/package.rs:586
``` 

---

The lockfile should be updated from the edited manifest

asmello:
This was intentional but you're right, the manifest should always take precedence.

PR #83, opened 2023-09-26
Fix manifest precedence

asmello:
The manifest file takes precedence over the lockfile when specifying where to download an artefact.


asmello:
Fixes #76

src/command.rs
```diff
@@ -93,9 +93,13 @@ pub async fn add(registry: RegistryUri, dependency: &str) -> eyre::Result<()> {
...
 
-    manifest
-        .dependencies
-        .push(Dependency::new(registry, repository, package, version));
+    manifest.dependencies.push(Dependency::new(
+        registry,
+        repository,
+        package,
+        version,
+        crate::manifest::DependencyType::Root,
```
mara-schulke, line 101: root is not serialized i guess?
why is it needed?

asmello: It's not needed here, it's used to build the dependency graph while allowing manifest-derived dependencies to override the lockfile metadata, but I haven't thought about serialisation so I'll confirm if that works sensibly.

mara-schulke:
I feel like we can undo the "dependencytype" and still solve the manifest precedence?

src/command.rs
asmello: Ok, just checked and yes it gets ignored as it's supposed to.

asmello:
> I feel like we can undo the "dependencytype" and still solve the manifest precedence?

Hmm, that was my initial idea but it felt awkward to carry extra dependency metadata around separate of the dependency struct, but since this is only really relevant in one place I think that might be best after all. Will look into how that looks like in practice.

src/command.rs
asmello: Removed it after all. 🤷

src/manifest.rs
```diff
@@ -178,6 +178,20 @@ impl Dependency {
             },
         }
     }
+
+    pub fn with_version(&self, version: &Version) -> Dependency {
```
mara-schulke, line 182: Same here

src/package.rs
```diff
@@ -584,15 +592,23 @@ impl DependencyGraph {
...
+            let package = registry
+                .download(dependency.with_version(&local_locked.version))
+                .await
+                .wrap_err_with(|| {
+                    format!(
+                        "Error downloading package {}@{}",
+                        local_locked.name, local_locked.version
+                    )
+                })?;
             local_locked.validate(&package).wrap_err_with(|| {
```
mara-schulke, line 612:
```suggestion

            local_locked.validate(&package).wrap_err_with(|| {
```

mara-schulke (approved):
Approved, but when reading through the algorithm it seems very complex! Maybe we could chat the next days on how to simplify it eventually

asmello:
> Approved, but when reading through the algorithm (not written in this pr tho!) it seems very complex! Maybe we could chat the next days on how to simplify it eventually

I'm skeptical it can be made any simpler (I tried hard to make it as simple as possible for the V0). I actually have ideas for a slightly more complex but much more intelligent version using a trick similar to what Poetry does. But let's chat, I'm not planning on pursuing any optimisations in the immediate future but we should align on vision.

PR #84, opened 2023-09-26
Add proxy and root cert docs

asmello:


docs/src/faq.md
```diff
@@ -34,3 +34,11 @@ token = "some-other-token"
 ## Why can't I log in with a username?
 
 `buffrs login` no longer supports the `--username` flag, as we no longer use BasicAuth. Instead we set the `X-JFrog-Art-Api` header.
+
+## How do I use a custom root certificate?
```
mara-schulke, line 38: I think this is not useful in the faq?

asmello: It's the kind of question I often see in FAQs, don't think it's bad to have a link here to the other page (which may not be super discoverable).

asmello: Ok, let's wait 'till there's evidence people can't find the other page. Removed from FAQ.

mara-schulke (approved):

PR #87, opened 2023-09-26
Instantiate a single registry instance per URI

asmello:
This refactor introduces the RegistryProvider struct which is responsible for lazily instantiating registries on a per-URI basis. This is useful to maximise usage of the connection pools, particularly when constructing the dependency graph.

This also abstracts away some of the code that was explicitly instantiating the Artifactory registry to use the Registry trait as intended.


asmello:
Should be merged after #83

src/command.rs
```diff
@@ -279,15 +284,19 @@ pub async fn login(mut credentials: Credentials, registry: RegistryUri) -> eyre:
         raw.trim().into()
     };
 
-    credentials.registry_tokens.insert(registry.clone(), token);
+    credentials.registry_tokens.insert(uri.clone(), token);
 
-    let artifactory = Artifactory::new(registry, &credentials)?;
+    match uri.kind() {
+        RegistryType::Artifactory => {
```
mara-schulke, line 290: Why do we need this?

src/manifest.rs
```diff
@@ -178,6 +178,20 @@ impl Dependency {
             },
         }
     }
+
+    pub fn with_version(&self, version: &Version) -> Dependency {
```
mara-schulke, line 182: Doc comment is missing

src/command.rs
asmello: We don't strictly need it, but in the future there will be other kinds of registry. We shouldn't assume it's always Artifactory.

src/registry/mod.rs
```diff
@@ -108,4 +109,41 @@ impl RegistryUri {
...
+    /// Returns the type of registry this URI refers to
+    pub fn kind(&self) -> RegistryType {
+        RegistryType::Artifactory
+    }
+}
+
+#[derive(Default)]
+pub struct RegistryProvider {
+    credentials: Credentials,
+    registries: HashMap<RegistryUri, Arc<dyn Registry + Send + Sync>>,
```
mara-schulke, line 122: this is overly complex, as we just have artifactory currently?

and if we move away from artifactory, we will have _exactly_ one network registry impl (the buffrs registry)

mara-schulke:
```suggestion
    registries: HashMap<RegistryUri, Arc<R>>,
```

src/registry/mod.rs
```diff
@@ -108,4 +109,41 @@ impl RegistryUri {
...
+    pub fn new(credentials: Credentials) -> Self {
+        Self {
+            credentials,
+            registries: Default::default(),
+        }
+    }
+
+    pub fn get_or_create(
+        &mut self,
+        registry_uri: &RegistryUri,
```
mara-schulke, line 135:
```suggestion
        uri: &RegistryUri,
```

(to keep this the same as everywhere else)

src/registry/mod.rs
```diff
@@ -108,4 +109,41 @@ impl RegistryUri {
 
         Ok(())
     }
+
+    /// Returns the type of registry this URI refers to
+    pub fn kind(&self) -> RegistryType {
```
mara-schulke, line 114: imo this can just go completely

src/registry/mod.rs
```diff
@@ -108,4 +109,41 @@ impl RegistryUri {
...
+        RegistryType::Artifactory
+    }
+}
+
+#[derive(Default)]
+pub struct RegistryProvider {
```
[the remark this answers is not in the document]
asmello: That doesn't work because we want this to hold more than one registry type (each URI could have a different registry type).

src/registry/mod.rs
mara-schulke: I know, but my suggested change is inline with what you wrote, only narrowing the solution to we "just have one registry backend" axiom

src/command.rs
mara-schulke: > but in the future there will be other kinds of registry
No 😅 
There will be: now artifactory, soon the buffrs registry (which entirely replaces artifactory

src/registry/mod.rs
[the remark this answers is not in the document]
mara-schulke: See my other comments, we wont!

src/registry/mod.rs
asmello: It can't, as we need a way to map URIs to the correct Registry implementations when instancing them. This is mostly relevant for RegistryProvider, and is one of the major benefits of implementing that struct, as it hides away the complexity of having multiple registry types.

asmello:
> General note on this: This seems like over optimization as we have never created more than a handful of clients atm? The implementation looks good, but is there any reason why we need this?
> 
> I think adoption + the grpc based registry are more important in terms of priority

I strongly disagree. The optimisation aspect of it is nice, but the main goal here is actually to refactor the code to abstract away instantiating different registry implementations directly. Implementing this now is important to lay a good foundation for future changes, so that we veer away from hardcoding Artifactory everywhere.

mara-schulke:
So, the underlying axiom of my comments is that we just do not support any other registry impl for now, and we will adapt this code if we should ever break that

The planned pipeline is:

```
Stabilization -> Buffrs Registry Implementation -> Buffrs Registry Migration (away from artifactory)
```

asmello:
> So, the underlying axiom of my comments is that we just do not support any other registry impl for now, and we will adapt this code if we should ever break that
> 
> The planned pipeline is:
> 
> ```
> Stabilization -> Buffrs Registry Implementation -> Buffrs Registry Migration (away from artifactory)
> ```

Yeah, I know that's not on the radar for now, but my concern is that we would end up stabilising on an API that hardcodes Artifactory, and it would be a pain to migrate out of it in the future. So ultimate goal here is to get the API right.

src/command.rs
asmello: Ok, this changes things! We need a different kind of refactor then. Will close this out.

PR #88, opened 2023-09-27
Improve error when attempting to publish impl packages

asmello:


src/package.rs
```diff
@@ -130,9 +130,12 @@ impl PackageStore {
     pub async fn release() -> eyre::Result<Package> {
         let manifest = Manifest::read().await?;
 
-        let pkg_manifest = manifest.package.clone();
+        ensure!(
+            manifest.package.kind != PackageType::Impl,
+            "Impl packages cannot be published"
```
mara-schulke, line 135:
```suggestion
            "Packages with type `impl` cannot be published"
```

mara-schulke (approved):

PR #90, opened 2023-09-27
Improve error when unauthorized

### LINKED ISSUE #75: Buffrs install gives a bad error message
```
λ buffrs install
Error:
   0: Server response has incorrect mime type: "application/json;charset=ISO-8859-1"

Location:
   src/registry/artifactory.rs:154

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
```

this happened after updating from buffrs v0.5 (with old credentials present) to buffrs v0.6

asmello:
Fixes #75

### DIFF
--- a/src/registry/artifactory.rs
+++ b/src/registry/artifactory.rs
@@ -128,9 +128,20 @@ impl Registry for Artifactory {
         let response = request.send().await?;
 
         ensure!(
-            response.status() != 302,
+            !response.status().is_redirection(),
             "Remote server attempted to redirect request - is this registry URL valid? {}",
             dependency.manifest.registry,
+        );
+
+        ensure!(
+            response.status() != 401,
+            "Unauthorized - please provide registry credentials with `buffrs login`",
+        );
+
+        ensure!(
+            response.status().is_success(),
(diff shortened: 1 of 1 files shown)


tetigi (approved):

src/registry/artifactory.rs
```diff
@@ -143,12 +154,6 @@ impl Registry for Artifactory {
             "Server response has incorrect mime type: {content_type:?}"
```
mara-schulke, line 154: If we have an invalid mimetype, can we print debug information in the eyre stacktrace?

Currently the mimetype error is a little unhelpful + we should only verify the mimetype when the response code is exactly 200!

src/registry/artifactory.rs
```diff
@@ -143,12 +154,6 @@ impl Registry for Artifactory {
...
         );
 
-        ensure!(
```
mara-schulke, line 146:
```
     ensure!(
            response.status().is_success(),
            "Failed to fetch {dependency}: {}",
            response.status()
        );
```

I do think this is a good idea?

PR #92, opened 2023-09-27
Better prompt for token in login command

asmello:
Screenshot 2023-09-27 at 15 02 40

### DIFF
--- a/src/registry/artifactory.rs
+++ b/src/registry/artifactory.rs
@@ -53,13 +53,7 @@ impl Artifactory {
 
         let response = request.send().await?;
 
-        let status = response.status();
-
-        if !status.is_success() {
-            tracing::info!("Artifactory response header: {:?}", response);
-        }
-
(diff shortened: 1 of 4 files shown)


src/registry/artifactory.rs
```diff
@@ -53,13 +53,7 @@ impl Artifactory {
...
         let response = request.send().await?;
 
-        let status = response.status();
-
-        if !status.is_success() {
-            tracing::info!("Artifactory response header: {:?}", response);
-        }
-
-        ensure!(status.is_success(), "Failed to ping artifactory");
+        response.error_for_status()?;
```
mara-schulke, line 56: Can we wrap this with a message?

mara-schulke (approved):
This does work with stdin-only right? (Talking about `echo x | buffrs login ..`}

asmello:
Depending on TTY manipulation has a number of downsides, I'll explore alternatives.

PR #95, opened 2023-09-29
Revamp errors

asmello:
Introducing `thiserror` for producing granular error types everywhere. This is primarily intended to support consuming Buffrs as a library, but also lays the foundations for easily producing better error messages (think `miette` and the likes).

There's A LOT of code touched by this PR, but the changes should be pretty straightforward. I tried not to make structural changes but some refactoring was unavoidable.


src/credentials.rs
```diff
@@ -31,55 +31,110 @@ pub struct Credentials {
     pub registry_tokens: HashMap<RegistryUri, String>,
 }
 
-impl Credentials {
-    fn location() -> eyre::Result<PathBuf> {
-        let home = home::home_dir().wrap_err("Failed to locate home directory")?;
+/// Error produced when locating the credentials file
+#[derive(Error, Debug)]
+pub enum LocateError {
```
j-baker, line 36: broad comment on everything. I do not like this PR. You've made a _tonne_ of implementation details public, which dramatically increases the surface area of this project.

I strongly, strongly, strongly recommend not adding specific error types for things you do not expect the user to be able to handle.

Why? Because each thiserror error is part of your public API, meaning that when you 1.0, adding a new type of error is a breaking API change.

So e.g. for reading credentials...

```
- Locate
- Io
- Toml
```

what is an API caller of this thing supposed to be doing if a 'locate' error comes back? Why are they supposed to differentiate between 'Io' and 'Toml'? But if you add a new mechanism to e.g. indicate that it's in a deprecated format, you've just made a breaking API change for the person who wants to process the fact that the credentials are missing. But what are they supposed to do about that?

mara-schulke: @j-baker general purpose behind this mr is to increase error messages to users! Thats why we moved away from eyre to thiserror.

We should unify the error enums where possible imo. so there should be one `buffrs::Error` wrapper which internally holds our typed errors, which keeps the error easy for users and enables good diagnostics at the same time

src/errors.rs
```diff
@@ -0,0 +1,19 @@
...
+#[derive(Error, Debug)]
+#[error("{msg}. Cause: {source}")]
+pub struct IoError {
+    source: std::io::Error,
+    msg: String,
+}
+
+impl IoError {
+    /// Constructs an IoError from an std IO error and a context message
+    pub fn new(source: std::io::Error, msg: impl Into<String>) -> Self {
```
mara-schulke, line 13: should this be beyond `pub(crate)`?

src/credentials.rs
mara-schulke: My proposal for this would be to make the error types internal only and returns a `buffrs::Error` type everywhere, i do think this level is good for diagnostics but at the moment it looks a bit cumbersome for users + for maintainers

src/credentials.rs
```diff
@@ -31,55 +31,110 @@ pub struct Credentials {
...
+    Io(std::io::Error),
+    /// Could not parse the TOML credentials data
+    #[error("Failed to deserialize credentials. Cause: {0}")]
+    Toml(toml::de::Error),
+}
+
+/// Error produced when updating the credentials file
+#[derive(Error, Debug)]
+pub enum WriteError {
+    /// Could not locate the credentials file
```
j-baker, line 73: these comments are totally redundant given the error itself contains the same information. The duplicative information simply increases the likelihood that it gets stale.

src/credentials.rs
mara-schulke: Of course we did, that is what we are currently using! The error messages which are displayed to users turn out to be not really helpful sadly, thats why we considered transitioning to a combination of miette and thiserror

mara-schulke: Having an "internal" error kind / error enum is pretty common tho. The io module of the std lib does this and offers both a `std::io::Error` + detailed diagnostics on the actual thing that went wrong.

asmello: Should probably have kept this one in draft state, as it really is still a draft! I figured I'd need significant changes.

> My proposal for this would be to make the error types internal only and returns a buffrs::Error type everywhere

+1, that makes a lot of sense, I think this is a pattern many libraries use (like `reqwest`, though I have some pet peeves with some decisions they made).

> Having an "internal" error kind / error enum is pretty common tho. The io module of the std lib does this and offers both a std::io::Error + detailed diagnostics on the actual thing that went wrong.

Also +1 on this, it's very useful to be able to access the underlying category of error, in case you want to customise error messages, write tests more robustly, etc. The end user doesn't really care about the typed error, but again this is for the sake of the library users that may compose Buffrs into other tools.

Overall, I tried to keep a balance between informative errors and keeping internal details out. Enough to distinguish a TOML syntax error from a file permission error, for example, because those two problems have very distinct solutions. That said, `thiserror` by nature does encourage more granular errors so happy to merge variants where it makes sense, too. Since I was doing this en masse I more or less created a distinct error variant for each use of `?` throughout the code, except where it was obvious it was the same error case in the same function.

In fact, I'm curious what error classes you two think would be useful to expose publicly.

src/errors.rs
asmello: I need to give this one more thought. The goal of `thiserror` is to move all the static details of an Error type to its struct declaration. That means this struct goes against `thiserror`'s philosophy, as it captures the context inline and makes it opaque to the wrapper struct. So when I wrote this I wasn't sure if I was doing a good thing.

That said, looking at this from a consumer's perspective, they probably don't care exactly _how_ an IO error was produced, just that it's an IO error, and we're capturing the context in the message field. So maybe I was actually on the right track and should do more of this.

src/credentials.rs
asmello: I agree. But I couldn't figure out a way to turn on the doc-comments linter without it complaining about the lack of these. I wish we could generate these from the field macro or vice-versa.

src/credentials.rs
```diff
@@ -31,55 +31,110 @@ pub struct Credentials {
...
+    #[error("IO error: {0}")]
+    Io(std::io::Error),
+    /// Could not parse the TOML credentials data
+    #[error("Failed to deserialize credentials. Cause: {0}")]
+    Toml(toml::de::Error),
+}
+
+/// Error produced when updating the credentials file
+#[derive(Error, Debug)]
+pub enum WriteError {
```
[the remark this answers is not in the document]
asmello: Which ones? All `WriteError` structs? There are slight differences, but maybe we can take a union of fields.

asmello:
Thanks for the initial feedback @mara-schulke and @j-baker! This is what I'm going to do next:
- [ ] Add a `buffrs::Error` struct that will be the single error type exported from this crate.
- [ ] Make all other errors `pub(crate)`.

The root Error type will have a `kind()` method that returns an enumeration that can be used to distinguish **categories** of errors that are shared globally across this crate. It will also always be Display as a short one liner, but to expose more context we can provide a `source()` method that will return an `impl Error` implemented by the original granular error. Since this generic only exposes the thin Error API it doesn't leak any implementation details, while still making it possible for the granular errors to provide rich contextualised messages.

Open questions:
- What categories of errors should we include in the `kind()` enumeration?
- How granular should the internal errors be? Which ones can we merge?
- Is IoError useful?
- Should we keep `#![warn(missing_docs)]` on or is this demanding too many useless comments?
  - Actually this is probably a moot point since I don't think it'll complain once we make the errors crate visible.

src/credentials.rs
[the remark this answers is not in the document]
mara-schulke: `WriteError` `ReadError` etc – imo they can all be joined to either one `CredentialsError` or even something like `FileError` which could be reused across the manifest, credentials, config impls etc. I dont have a strong opinion on what is better, but i think having one error enum per function is to granular

src/errors.rs
mara-schulke: I think we should have `thiserror` enums and capture them in an `eyre::Report`-like structure which can keep track of traces. So i imagine there can be multiple errors which affect eachother that should be stacked / traced. Eg. when trying to install a dependency and it fails due to missing credentials i imagine something like this:

```
1. InstallationError
2. AuthenticationError
3. CredentialsError```

I dont know whether using an eyre report or a custom impl is better – the benefit of implementing this would be the detailed control over how it is printed in the end + the typesafety

File: src/credentials.rs (hunk 3 above)
Comment (mara-schulke) [reply src/credentials.rs:73]: i dont think its _that_ bad to keep them. it is redundant in code, but nice for docs.rs + users in their ide

**Comment by mara-schulke** [comment]
Your action plan sounds good @asmello 😊

**Comment by asmello** [comment]
I've been doing some further reading on error handling best practices. I think we're on the right track with the dual internally-granular and externally-opaque principles.

That said there are some additional things I've been thinking about:
- **Composition:** [source errors should not be directly included when the wrapper error is displayed](https://kazlauskas.me/entries/errors#:~:text=Never%20format%20the%20er%C2%ADror%20source%20into%20the%20dis%C2%ADplayed%20mes%C2%ADsage.%20Users%20can%20use%20Error%3A%3Asource%20to%20ob%C2%ADtain%20the%20in%C2%ADform%C2%ADa%C2%ADtion%20in%20their%20present%C2%ADa%C2%ADtion%20lo%C2%ADgic.). We should delegate  to libraries like `eyre` and `miette` that can render error chains nicely.
- I do think having function-specific error types is a good thing, because while there's some overlap between errors produced by functions in each module, we want their returned enums to be as specific as possible. Any variants that can never be returned from a function shouldn't be in the enum! I don't want to handle TOML parsing failure when calling a function that only checks the existence of a file. Also it hurts context capture, as we lose information about where the error was produced, and modularisation, as it entangles functions to co-depend on a shared type. This is covered extensively in [this writeup](https://sabrinajewson.org/blog/errors) that I really like.
- I think we should make all error messages lowercase, to be in line [with core library recommendations](https://doc.rust-lang.org/stable/std/error/trait.Error.html#:~:text=Error%20messages%20are%20typically%20concise%20lowercase%20sentences%20without%20trailing%20punctuation%3A).

So in addition to the previous plan, I will also work on:
- [ ] Change the error messages to be lowercase one liners that don't include source messages.
- [ ] Include source errors in some places where I've not included it out of fear of creating a hard dependency on 3rd-party libraries. Since these granular errors won't be public, this isn't a concern anymore and we get the benefit of even more specific error messages via chaining.

File: src/credentials.rs (hunk 3 above)
Comment (asmello) [reply src/credentials.rs:73]: I think this becomes moot since we'll make these types crate private. I might remove some trivial comments but I guess there's some value in duplicating the message for the sake of IDE integration for collaborators, still.

File: src/credentials.rs (hunk 4 above)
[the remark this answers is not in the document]
Comment (asmello) [reply src/credentials.rs:72]: See [my recent comment](https://github.com/helsing-ai/buffrs/pull/95#issuecomment-1741741985) on why I think this kind of merging is a bad idea. We could also have a wrapper type at the module level but I think that is overkill. I may still do some merging across variants, but I'm now skeptical about merging errors altogether.

In fact there may be some benefit to defining unique types even if there's exactly one failure mode, in order to support future extensibility in case the implementation becomes more complex (more specifically, functions like `Lockfile::exists()` that currently return `IoError` would return their own local type that works like `IoError` but is marked as `#[non_exhaustive]` so that it could be extended without a break). Not 100% on this yet but that's my inclination.

File: src/credentials.rs (hunk 3 above)
Comment (asmello) [reply src/credentials.rs:73]: I found this nice crate: [displaydoc](https://docs.rs/displaydoc/latest/displaydoc/). It works well with `thiserror` so I'll be converting some doc-comments to this.

**Comment by asmello** [comment]
Worked on changes today (haven't pushed yet), and as I went through the code again I started to think that granular public errors are **not** the root of the problem, and not intrinsically a bad thing. The issue James flagged with leaking internal details is extremely valid, but it doesn't mean we should drop error enumeration altogether. I believe it is a fallacy that we should use `std::io::Error` as a role model (more on that later).

Here are the actual things we wish to avoid:
- Leaking types from 3rd-party libraries via our public API
- Exposing details that are subject to change in the future
- Creating friction to adding new error conditions (or metadata) — in other words, prevent the library from evolving

To implement a good error system that can evolve fearlessly, it should be sufficient to:
- Hide 3rd-party types as private fields and provide accessor methods with stable signature as an abstraction layer
  - All the exposed types should be defined locally in the crate or be part of `std`
- Merge similar or related error conditions together, and capture implementation-defined errors under an abstract `Other` variant or equivalent
- Mark enums as `#[non_exhaustive]` so that variants can continue to be added (same for the fields)

Taking this example from James' comment earlier:

> So e.g. for reading credentials...
> 
> - Locate
> - Io
> - Toml
> 
> what is an API caller of this thing supposed to be doing if a 'locate' error comes back? 

A Locate error here (actually `Locate::MissingHome`) would indicate that the environment is misconfigured. Since this is an environmental problem, there usually isn't much the caller function can do to handle the error itself, but suppose the caller was a tool that manages virtual environments and initialises a sandboxed Buffrs project, then perhaps this would be a recoverable error (not that wild of an example! this is something that could potentially happen in a tests module). At the very least, by capturing this failure mode with a unique variant, we allow the caller to safely remap the error to a local type, change the display message etc. — this is vital for healthy composition across crates.

Consider if we instead just mapped this failure mode to an Io variant, we'd have to resort to string matching to distinguish between environment problems and filesystem errors, which have very different solutions.

It's easy for a lib to drop type information, impossible for consumers to safely recover it.

Another thing, we don't want to directly expose the `toml::de::Error` error inside the Toml variant publicly, but that doesn't mean a parsing-related variant isn't important to have. It just needs to abstract away the specifics of how we parse the payload.

BTW I like [this writeup](https://mmapped.blog/posts/12-rust-error-handling.html) I found today, it covers a good chunk of good practices.

## Why I don't think `std::io::Error` is a good role model

The IO module faces the relatively peculiar problem of having to report heavily platform dependent errors. Not only there are distinct sets of errors that can happen for the same operation as implemented in different runtimes, the details available about those errors aren't particularly consistent and are subject to change outside of Rust's version control ecosystem. Thus to be portable, the exposed Error type must only capture the common denominator and delegate platform-dependent specialisation to business logic.

This sort of situation can happen in crates that must interface with some types of external low-level libraries, and that are foundational for a wide array of applications. But Buffrs has the privilege of only interacting with other Rust libraries and its direct and transitive error conditions are all well defined. So there's no reason for its errors types to be agnostic to the underlying context.

Moreover, there _are_ downsides to the single-public-error approach worth noting:
- To recover the fully contextualised errors one must downcast, which is an unsafe operation
- The caller can only rely on documentation (and source-code inspection) to determine which conditions can lead to error results
  - This is often inadequate! Consider how `reqwest::RequestBuilder::build()` can fail, but docs don't mention how. The provided Error variant also doesn't give us a clue, as the type is used for anything from TLS misconfiguration, connection errors or invalid HTTP header value... We have no way of knowing and can only hope to get an informative display message!
  - `std::io::Error::kind()` mitigates this problem a bit, but it's still a problem that we have no reliable way of knowing which variants can return from a given call...

I think that as long as we account for the pitfalls mentioned earlier in this comment, defining a few high-signal public variants is a sensible design for Buffrs. In particular, I think it's ok for each function in the `command` module to define its own error type, as they have well defined behaviour with well defined failure modes. Outside of that, a few common errors defined globally should probably cover the rest - HttpError, ParseError, IoError etc. each one abstracting away the specifics of 3rd-party libraries and how they are used, of course. Here's one example:

```rust
[derive(thiserror::Error, Debug)]
#[error("{path}: at line {line} column {column}: {message}")]
pub struct ParseError {
    path: PathBuf,
    line: usize,
    column: usize,
    message: String,
}

impl ParseError {
    pub fn path(&self) -> &Path {
        &self.path
    }

    pub fn line(&self) -> usize {
        self.line
    }

    pub fn column(&self) -> usize {
        self.column
    }

    pub fn message(&self) -> &str {
        &self.message
    }
}
```

There might be more context to be captured, but this is the basic idea of something that doesn't expose any unnecessary internal details, can compose nicely and should integrate well with `miette` and other reporting crates.

Another tentative example I've been fiddling with:

```rust
/// Error produced when initializing a Buffrs project
#[derive(Error, Display, Debug)]
#[non_exhaustive]
pub enum InitError {
    /// a manifest file was found, project is already initialized
    AlreadyInitialized,
    /// invalid package name
    InvalidName(#[from] PackageNameValidationError),
    /// IO error
    Io(#[from] std::io::Error)
}
```

Unless we make a breaking change to the `init` command (for example, by changing its default behaviour to override the manifest file if it exists), this interface is stable. Even then, there are cases where we may simply deprecate variants by making them unused until the next major version. Even though this error type is tied to a single function, it shouldn't create any maintenance friction.

asmello:
This PR is ready for another round of feedback.

src/credentials.rs
```diff
@@ -31,55 +35,99 @@ pub struct Credentials {
...
+pub enum LocateError {
+    /// the home folder could not be determined
+    MissingHome,
+}
+
+#[derive(Error, Display, Debug)]
+#[non_exhaustive]
+#[allow(missing_docs)]
+pub enum ExistsError {
+    /// could not locate the credentials file
```
mara-schulke, line 50: Can we upercase the doc comments, and lowercase them manually in the wrapping error?

asmello: The doc comments are now unified with the error messages, so unfortunately this stylistic conflict is something we have to live with. The alternative is to have duplicative comments and messages like before.

src/credentials.rs
```diff
@@ -31,55 +35,99 @@ pub struct Credentials {
...
+    Locate(#[from] LocateError),
+    /// could not read the file
+    Io(#[from] std::io::Error),
+    /// could not deserialize the credentials
+    Deserialize(#[from] DeserializationError),
+}
+
+#[derive(Error, Display, Debug)]
+#[non_exhaustive]
+#[allow(missing_docs)]
```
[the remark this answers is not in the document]
asmello: Unfortunately not, because if we add a top-level doc-comment it will override the per-variant display implementation. It's another downside of using displaydoc...

src/command.rs
```diff
@@ -26,28 +27,57 @@ use crate::{generator, generator::Language};
...
+    /// io error: {message}
+    Io {
+        message: String,
+        source: std::io::Error,
+    },
+    /// failed to create manifest file
+    ManifestWrite(#[from] manifest::WriteError),
+    /// failed to create package store
+    PackageStore(#[from] package::CreateStoreError),
+    /// {0}
```
mara-schulke, line 54: is this a meaningful doc comment

asmello: Not at all. It's just there to tell displaydoc how to render the Other variant (ie, transparently). I could use `#[displaydoc("...")]` to make it possible to additionally include a non-display doc-comment, but I'm not sure what I'd put there.

asmello:
@jonhoo this is a complex and subtle enough design decision that we'd deeply value your input as well!

mara-schulke:
Lets not close this already

asmello:
Superseded by #103

PR #97, opened 2023-10-02
Use the git feature flag

ntrinquier:
Since the code block is using `git2`, and that otherwise the `git` feature is unused, the code block should be behind the `git` flag, not `build`?


src/command.rs
```diff
@@ -149,7 +149,7 @@ pub async fn publish(
     #[cfg(feature = "build")] allow_dirty: bool,
```
mara-schulke, line 149: This has to be the git feature aswell then!

ntrinquier: Indeed, changed!

mara-schulke (approved):

mara-schulke:
Ah, i think you have to update the command struct aswell!

mara-schulke:
(Talking about the `src/main.rs`)

PR #98, opened 2023-10-02
add flake.nix

koiuo:
This PR makes _buffrs_ available as a [Nix](https://nixos.org/guides/how-nix-works) [flake](https://nixos.wiki/wiki/Flakes).

After this is merged _buffrs_ can be installed with

```
nix profile install github:helsing-ai/buffrs
```

(in the meantime, this can be installed from my fork `nix profile install github:koiuo/buffrs`)

Apart from making installation easier, this provides a simple development shell with all _buffrs_ development dependencies automatically installed

```
nix develop
```

This has only been tested on Linux x86_64, but theoretically should work on any other Linux-based platform where Nix is available.

This has NOT been tested on Mac OS (nix-darwin).


mara-schulke:
Hi @koiuo thank for your contribution! 😊

flake.nix
```diff
@@ -0,0 +1,32 @@
+{
+  inputs = {
+    flake-utils.url = "github:numtide/flake-utils";
+    naersk.url = "github:nix-community/naersk";
```
j-baker, line 4: Hi! from your perspective, how does Naersk compare with Crane (if you've used both?) in terms of UX & philosophy? We're familiar with Crane but not with naersk as of present. We do some Nix building internally and so ideally would become most familiar with one Rust build tool.

flake.nix
```diff
@@ -0,0 +1,32 @@
...
+  outputs = { self, flake-utils, naersk, nixpkgs }:
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        pkgs = (import nixpkgs) {
+          inherit system;
+        };
+
+        naersk' = pkgs.callPackage naersk {};
+      in rec {
+        defaultPackage = naersk'.buildPackage {
```
j-baker, line 17: realistically, this will break in time unless it's somehow tested in CI. Any chance you could add a gitlab action which runs nix flake check or similar (and ensure that defaultPackage is built in this case)?

Could use https://github.com/DeterminateSystems/magic-nix-cache-action to ensure that the rebuild cost is low...

flake.nix
koiuo: I haven't used crane. Naersk is recommended by the NixOS wiki and that's the only thing I tried.

I tried it on a more complicated repo, with multiple rust modules.

koiuo:
Hey. Thanks a lot for the initial review.

I'm up for addressing all the comments, however I'm off w/o access to my dev machine for another two weeks. Feel free to close this if anyone else is eager to contribute while I'm off.

j-baker:
I've made this MR: https://github.com/helsing-ai/buffrs/pull/104 which:
- rebases this
- adds CI
- removes some deprecations (defaultPackage vs packages.default, devShell vs devShells.default)
- makes sure the main build is run in CI.

Some point later I will refactor the MR to use crane. Among other things:
- Crane encourages running various cargo commands as first class things, so we can likely collapse other parts of the CI process and just run nix checks for CI
- It's what we use internally so knowledge is better.

mara-schulke:
Superseded by #104 😊 

Let me know if #104 covers your use case. If not feel free to file an issue! @koiuo

Issue #89, opened 2023-09-27
Simplify structure of the registry module


asmello:
We plan to only ever support one Registry implementation. Currently it's Artifactory, and in the future it will be our own implementation. But at no point we plan to support multiple types of registry.

The `registry` module is structured to provide a Registry trait, which was originally introduced so that multiple registry types could co-exist in the codebase. Since this will never happen, we don't need that trait, and in fact it existing can only lead to confusion about our intent.

In order to simplify the codebase and make it easier for new contributors to understand it, we should remove the Registry trait and refactor the `registry` module.

PR #99, opened 2023-10-03
Remove Registy trait in registry module and refactor

MikP0:
This PR removes the `Registry`  trait from the registry module - according to #89 . The following changes has been made:
1. Removed `Registry` trait and `RegistryType` enum from `registry/mod.ts`
2. Moved Artifactory functions that were implemented for the `Registry` trait into the struct implementation - added made them `pub`(lic) 
3. Moved `LocalRegistry` functions that were implemented for the `Registry` trait into the struct implementation - added made them `pub`(lic) 
4. Removed `Registry` trait references


asmello:
I've been thinking about ways to test the registry interactions, and it occurred to me that even though we don't want to support multiple registry implementations, the trait might still be helpful so we can inject a local registry for use in testing.

That may still require a larger refactor though (because the current signatures don't give us a way to customise the registry choice) so I think it may still be best to get rid of the trait for now and reintroduce it later when necessary. It's definitely confusing to have it at the moment.

asmello (approved):

asmello:
Thanks for the contribution!

PR #103, opened 2023-10-06
Improve Errors using `miette`

asmello:
The primary goal of this PR is to implement some error reporting improvements developed for #95 while avoiding introducing new public types.

### What's different from #95
Instead of introducing typed errors everywhere, I made judicious use of them **only** for error reusability (so that structurally similar errors need only be defined once).

This approach preserves the current pattern of defining opaque errors inline and leaking `eyre::Report` as the crate's external error type. It does not introduce any new public error types.


src/command.rs
```diff
@@ -26,27 +26,28 @@ use crate::{generator, generator::Language};
...
+        eyre::bail!("a manifest file was found, project is already initialized");
+    }
 
     fn curr_dir_name() -> eyre::Result<PackageName> {
-        std::env::current_dir()?
+        std::env::current_dir()
+            .wrap_err("failed to access current directory")?
             .file_name()
-            .wrap_err(DIR_ERR)?
+            .expect("unexpected error: current directory path terminates in ..")
```
mara-schulke, line 48:
```suggestion
            .wrap_err("unexpected error: current directory path terminates in ..")?
```

src/command.rs
```diff
@@ -26,27 +26,28 @@ use crate::{generator, generator::Language};
...
             .to_str()
-            .wrap_err(DIR_ERR)?
+            .ok_or_else(|| eyre::eyre!("current directory path is not valid utf-8"))?
```
mara-schulke, line 50:
```suggestion
            .wrap_err("current directory path is not valid utf-8")?
```

src/command.rs
```diff
@@ -26,27 +26,28 @@ use crate::{generator, generator::Language};
 use std::path::PathBuf;
 
 use async_recursion::async_recursion;
-use eyre::{ensure, Context, ContextCompat};
+use eyre::Context;
```
mara-schulke, line 29:
```suggestion
use eyre::{Context, ContextCompat};
```

asmello: I think this is worse, `eyre` only provides `ContextCompat` for helping migrate existing `anyhow` projects, but [they recommend not using it in general](https://docs.rs/eyre/latest/eyre/#context-and-option).

src/command.rs
asmello: Related to the above (re: `eyre` deprecating option wrappers).

src/command.rs
```diff
@@ -62,80 +65,100 @@ pub async fn init(kind: PackageType, name: Option<PackageName>) -> eyre::Result<
...
 
     let (package, version) = dependency
         .split_once('@')
-        .wrap_err_with(|| format!("Invalid dependency specification: {dependency}"))?;
+        .ok_or_else(|| miette!("dependency specification is missing version part: {dependency}"))?;
 
     let package = package
         .parse::<PackageName>()
-        .wrap_err_with(|| format!("Invalid package name supplied: {package}"))?;
+        .wrap_err_with(|| format!("invalid package name: {package}"))?;
```
mara-schulke, line 100:
```suggestion
        .wrap_err_with(|| miette!("invalid package name: {package}"))?;
```

src/command.rs
```diff
@@ -62,80 +65,100 @@ pub async fn init(kind: PackageType, name: Option<PackageName>) -> eyre::Result<
...
 
     let version = version
         .parse::<VersionReq>()
-        .wrap_err_with(|| format!("Invalid version requirement supplied: {package}"))?;
+        .into_diagnostic()
+        .wrap_err_with(|| format!("not a valid version requirement: {version}"))?;
```
mara-schulke, line 105:
```suggestion
        .wrap_err_with(|| miette!("not a valid version requirement: {version}"))?;
```

src/command.rs
```diff
@@ -62,80 +65,100 @@ pub async fn init(kind: PackageType, name: Option<PackageName>) -> eyre::Result<
...
 
     manifest
         .dependencies
         .push(Dependency::new(registry, repository, package, version));
 
-    manifest.write().await
+    manifest
+        .write()
+        .await
+        .wrap_err("failed to write manifest file")
```
mara-schulke, line 118:
```suggestion
        .wrap_err_with(|| miette!("failed to write manifest file"))
```

(would be even better to add the helpful details here)

src/command.rs
```diff
@@ -62,80 +65,100 @@ pub async fn init(kind: PackageType, name: Option<PackageName>) -> eyre::Result<
...
-            "Unable to remove unknown dependency {package:?}"
-        ))?;
+        .ok_or_else(|| miette!("package {package} not in manifest"))?;
 
     let dependency = manifest.dependencies.remove(match_idx);
 
-    PackageStore::uninstall(&dependency.package).await.ok();
+    // if PackageStore::uninstall(&dependency.package).await.is_err() {
+    //     tracing::warn!("failed to uninstall package {}", dependency.package);
+    // }
```
mara-schulke, line 137: Why is this / can we comment this in again?

src/command.rs
```diff
@@ -62,80 +65,100 @@ pub async fn init(kind: PackageType, name: Option<PackageName>) -> eyre::Result<
...
 
-    manifest.write().await
+    PackageStore::uninstall(&dependency.package).await.ok(); // temporary due to broken test
```
mara-schulke, line 139: Ah. This is bad, can we fix it?

src/command.rs
```diff
@@ -62,80 +65,100 @@ pub async fn init(kind: PackageType, name: Option<PackageName>) -> eyre::Result<
...
+
+    manifest
+        .write()
+        .await
+        .wrap_err("failed to write manifest file")
```
mara-schulke, line 144: Same

src/lock.rs
```diff
@@ -294,9 +335,16 @@ impl Lockfile {
...
         };
 
-        fs::write(LOCKFILE, toml::to_string(&raw)?.into_bytes())
-            .await
-            .wrap_err("Failed to write lockfile")
+        fs::write(
+            LOCKFILE,
+            toml::to_string(&raw)
+                .into_diagnostic()
+                .wrap_err_with(|| SerializationError("lockfile"))?
```
mara-schulke, line 342: why is the one (WriteError) using a const and this is using a str?

src/manifest.rs
```diff
@@ -78,30 +82,53 @@ pub struct Manifest {
...
-        Ok(raw.into())
+    /// Loads the manifest from the given path
+    pub async fn read_from(path: impl AsRef<Path>) -> miette::Result<Self> {
+        let path_ref = path.as_ref();
+
+        match fs::read_to_string(&path_ref).await {
+            Ok(contents) => {
+                let raw: RawManifest = toml::from_str(&contents)
+                    .into_diagnostic()
+                    .wrap_err_with(|| DeserializationError("manifest"))?;
```
mara-schulke, line 105: same

src/package.rs
```diff
@@ -87,12 +101,23 @@ impl PackageStore {
 
         fs::create_dir_all(&pkg_dir)
             .await
-            .wrap_err("Failed to install dependencies")?;
+            .into_diagnostic()
+            .wrap_err_with(|| {
+                format!(
```
mara-schulke, line 106:
```suggestion
                miette!(
```

src/package.rs
```diff
@@ -87,12 +101,23 @@ impl PackageStore {
...
+            })?;
 
-        tar.unpack(pkg_dir.clone()).wrap_err(format!(
-            "Failed to unpack tar of {}",
-            package.manifest.package.name
-        ))?;
+        tar.unpack(pkg_dir.clone())
+            .into_diagnostic()
+            .wrap_err_with(|| {
+                format!(
```
mara-schulke, line 115:
```suggestion
                miette!(
```

src/package.rs
```diff
@@ -103,63 +128,58 @@ impl PackageStore {
...
+                "depending on API packages is not allowed for {} packages",
+                manifest.package.kind
             );
         }
 
         let pkg_path = fs::canonicalize(&Self::PROTO_PATH)
             .await
+            .into_diagnostic()
             .wrap_err_with(|| {
                 format!(
```
mara-schulke, line 181:
```suggestion
                miette!(
```

src/package.rs
```diff
@@ -179,53 +200,61 @@ impl PackageStore {
...
+                .wrap_err_with(|| format!("failed to open file {}", entry.display()))?;
 
             let mut header = tar::Header::new_gnu();
             header.set_mode(0o444);
             header.set_size(
                 file.metadata()
+                    .into_diagnostic()
                     .wrap_err_with(|| {
-                        format!("Failed to fetch metadata for entry {}", entry.display())
+                        format!("failed to fetch metadata for entry {}", entry.display())
```
mara-schulke, line 223:
```suggestion
                        miette!("failed to fetch metadata for entry {}", entry.display())
```

src/package.rs
```diff
@@ -305,20 +334,23 @@ impl Package {
...
-    fn try_from(tgz: Bytes) -> eyre::Result<Self> {
+    fn try_from(tgz: Bytes) -> Result<Self, Self::Error> {
         let mut tar = Vec::new();
 
         let mut gz = flate2::read::GzDecoder::new(tgz.clone().reader());
 
         gz.read_to_end(&mut tar)
-            .wrap_err("Failed to decompress package")?;
+            .into_diagnostic()
+            .wrap_err("failed to decompress package")?;
```
mara-schulke, line 346:
```suggestion
            .wrap_err_with(|| miette!("failed to decompress package"))?;
```

mara-schulke: Although i think using `.wrap_err(miette!(..))` would be more readable, it creates the error before it is actually ocuring but its much more readable and concise. I dont think we are in a memory critic context here so i would value maintainability

src/command.rs
asmello: > (would be even better to add the helpful details here)
trying to keep the PR size under control but agree

src/package.rs
```diff
@@ -590,48 +635,58 @@ impl DependencyGraph {
...
             Ok(package)
         } else {
-            let registry = Artifactory::new(dependency.manifest.registry.clone(), credentials)?;
-            registry.download(dependency).await
+            let registry = Artifactory::new(dependency.manifest.registry.clone(), credentials)
+                .wrap_err_with(|| DownloadError {
+                    name: dependency.package.clone(),
+                    version: dependency.manifest.version.clone(),
+                })?;
+            registry
```
mara-schulke, line 679:
```suggestion

            registry
```

src/package.rs
```diff
@@ -501,13 +537,20 @@ pub struct DependencyGraph {
     entries: HashMap<PackageName, ResolvedDependency>,
 }
 
+#[derive(Error, Debug)]
+#[error("failed to download dependency {name}@{version} from the registry")]
+struct DownloadError {
```
mara-schulke, line 542: can we miette-derive this?

src/command.rs
asmello: yeah but want to do that in a separate PR as it's outside of the scope of this one

src/registry/artifactory.rs
```diff
@@ -25,72 +29,99 @@ pub struct Artifactory {
...
         );
 
         let version = dependency
             .manifest
             .version
             .comparators
             .first()
-            .wrap_err("internal error")?;
+            // validated above
+            .expect("unexpected error: empty comparators vector in VersionReq");
```
mara-schulke, line 109: I have to say this buggs me a lot, people will make errors and then we will have a runtime panic.

mara-schulke: Can we please please not panic? 🙇‍♀️

src/registry/artifactory.rs
```diff
@@ -25,72 +29,99 @@ pub struct Artifactory {
...
-        };
-
-        let mut request = self.client.get(repositories_url);
+    async fn make_auth_request(
+        &self,
+        method: Method,
+        url: Url,
+        body: Option<Bytes>,
+    ) -> miette::Result<Response> {
+        let mut request_builder = self.client.request(method.clone(), url.clone());
```
mara-schulke, line 55:
```suggestion
        let mut request = self.client.request(method, &url);
```

does this work?

src/registry/artifactory.rs
```diff
@@ -25,72 +29,99 @@ pub struct Artifactory {
...
-    pub async fn ping(&self) -> eyre::Result<()> {
-        let repositories_url: Url = {
-            let mut uri = self.registry.to_owned();
-            let path = &format!("{}/api/repositories", uri.path());
-            uri.set_path(path);
-            uri.into()
-        };
-
-        let mut request = self.client.get(repositories_url);
+    async fn make_auth_request(
```
mara-schulke, line 49: I dont like this change, it creates hard ti track, oop-like patterns.

src/registry/local.rs
```diff
@@ -32,31 +38,35 @@ impl LocalRegistry {
...
         );
 
         let version = dependency
             .manifest
             .version
             .comparators
             .first()
-            .wrap_err("internal error")?;
+            // validated above
+            .expect("unexpected error: empty comparators vector in VersionReq");
```
mara-schulke, line 54: same

src/registry/local.rs
```diff
@@ -71,11 +81,17 @@ impl LocalRegistry {
...
 
-        let bytes = Bytes::from(std::fs::read(path)?);
+        let bytes = Bytes::from(
+            std::fs::read(path)
+                .into_diagnostic()
+                .wrap_err("could not read file")?,
+        );
+
         Package::try_from(bytes)
+            .wrap_err_with(|| format!("failed to download dependency {}", dependency.package))
```
mara-schulke, line 91:
```suggestion
            .wrap_err_with(|| miette!("failed to download dependency {}", dependency.package))
```

src/registry/mod.rs
```diff
@@ -57,39 +58,31 @@ impl Display for RegistryUri {
...
+    type Err = miette::Report;
 
     fn from_str(value: &str) -> Result<Self, Self::Err> {
-        let slf = Self(Url::from_str(value)?);
-        slf.sanity_check_url()?;
-        Ok(slf)
+        let url = Url::from_str(value)
+            .into_diagnostic()
+            .wrap_err_with(|| format!("not a valid URL: {value}"))?;
+        sanity_check_url(&url)?;
```
mara-schulke, line 67:
```suggestion

        sanity_check_url(&url)?;
        
```

src/command.rs
mara-schulke: Can you create a bug report?

src/registry/artifactory.rs
asmello: Refactor done.

mara-schulke: From experience, someone will remove the check, without removing the panic. I think its not controversial to say, never panic in a library or application below the main function

src/registry/artifactory.rs
asmello: I don't understand, all this does is move some common logic to a function so we don't need to repeat it a bunch of times.

src/registry/artifactory.rs
asmello: No, not with the reference, but the clones are not necessary (they're residual from PR #95) and I'll remove them.

src/command.rs
asmello: https://github.com/helsing-ai/buffrs/issues/115

src/command.rs
```diff
@@ -62,80 +65,100 @@ pub async fn init(kind: PackageType, name: Option<PackageName>) -> eyre::Result<
...
 }
 
 /// Removes a dependency from this project
-pub async fn remove(package: PackageName) -> eyre::Result<()> {
-    let mut manifest = Manifest::read().await?;
+pub async fn remove(package: PackageName) -> miette::Result<()> {
+    let mut manifest = Manifest::read()
+        .await
+        .wrap_err("failed to read manifest file")?;
```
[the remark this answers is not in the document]
asmello: I assume this about us having more context to add here? I could include the package name but the caller function already reports the package name so it'd be a bit redundant.

src/registry/artifactory.rs
mara-schulke: (Im not critical about the fact that we abstract common logic)

asmello: I'm not sure how that would look like (I don't have much experience with pure functional languages). Since the existing code is imperative, my bias would be to keep a trivial refactor here and file an issue for changing the paradigm later, but if you have a simple change in mind happy to incorporate it (I just don't know what it would be).

mara-schulke: So you can usually use data lenses to extract the data you want from the parent data type, to then detect whether you want to error or not.

While writing / thinking about it a second time i found that you dont really do data introspection, but rather error validation here, so the most idiomatic thing imo would be to use the type system as much as possible:

```rust
struct Validated(Response);

impl TryFrom<Response> for Validated {
    type Error = miette::Error;

    fn try_from(/**/) -> Self {
        // do the ensuring and miette stuff
        
        Self(res)
    }
}

impl TryFrom<Validated> for Package {
// ...
}

// using in the toplevel function that does IO
fn download(&self, dep: Dependency) -> miette::Result<()> {
    let request = // build the request
    let response = response.await?;
    let package: Package = Validated::try_from(response)?.try_into()?;
    Ok(package)
}

File: src/command.rs (hunk 1 above)
Comment (mara-schulke) [reply src/command.rs:46]: Can we do it for consistency? I know you explained to to me, your reasoning is valid, this is just a OCD nit

File: src/command.rs [hunk 26]
```diff
@@ -203,12 +210,14 @@ pub async fn install(credentials: Credentials) -> eyre::Result<()> {
         graph: &DependencyGraph,
         locked: &mut Vec<LockedPackage>,
         prefix: String,
-    ) -> eyre::Result<()> {
+    ) -> miette::Result<()> {
         let resolved = graph
             .get(name)
-            .wrap_err("Dependency missing from dependency tree")?;
+            .expect("unexpected error: missing dependency in dependency graph");
```
mara-schulke, line 216: Same, please dont expect anything

src/command.rs
```diff
@@ -250,58 +259,52 @@ pub async fn install(credentials: Credentials) -> eyre::Result<()> {
...
+            .wrap_err(miette!("failed to read the token from the user"))?;
 
         raw.trim().into()
     };
 
     credentials.registry_tokens.insert(registry.clone(), token);
 
-    let artifactory = Artifactory::new(registry, &credentials)?;
-
     if env::var("BUFFRS_TESTSUITE").is_err() {
```
mara-schulke, line 296: this should also be a const, you made the other var a const aswell

src/credentials.rs
```diff
@@ -31,55 +36,73 @@ pub struct Credentials {
...
 
     /// Writes the credentials to the file system
-    pub async fn write(&self) -> eyre::Result<()> {
+    pub async fn write(&self) -> miette::Result<()> {
         fs::create_dir(
-            Self::location()?
+            location()?
                 .parent()
-                .wrap_err("Invalid credentials location")?,
+                .expect("unexpected error: resolved credentials path has no parent"),
```
mara-schulke, line 83: no expect please

src/credentials.rs
```diff
@@ -31,55 +36,73 @@ pub struct Credentials {
...
         )
         .await
-        .ok();
+        .ok(); // ok to ignore if parent already exists (other failure modes covered by expect above)
```
mara-schulke, line 86: This is prone to change

src/generator.rs
```diff
@@ -45,17 +46,22 @@ pub enum Generator {
...
+    /// Tonic include file name
     pub const TONIC_INCLUDE_FILE: &str = "buffrs.rs";
 
     /// Run the generator for a dependency and output files at the provided path
-    pub async fn run(&self) -> eyre::Result<()> {
-        let protoc = protoc_bin_path().wrap_err("Unable to locate vendored protoc")?;
+    pub async fn run(&self) -> miette::Result<()> {
+        let protoc = protoc_bin_path()
+            .into_diagnostic()
+            .wrap_err("unable to locate vendored protoc")?;
```
mara-schulke, line 64: miette

src/lock.rs
```diff
@@ -186,7 +206,7 @@ impl LockedPackage {
     ) -> Self {
         let digest = digest::digest(&digest::SHA256, &package.tgz)
             .try_into()
-            .expect("Unexpected error: only SHA256 is supported");
+            .expect("unexpected error: only SHA256 is supported");
```
mara-schulke, line 209: please no expect

src/lock.rs
```diff
@@ -205,28 +225,44 @@ impl LockedPackage {
...
+    pub fn validate(&self, package: &Package) -> miette::Result<()> {
         let digest: Digest = digest::digest(&digest::SHA256, &package.tgz)
             .try_into()
             .unwrap();
 
+        #[derive(Error, Debug)]
+        #[error("{property} mismatch - expected {expected}, actual {actual}")]
+        struct ValidationError {
+            property: &'static str,
+            expected: String,
```
mara-schulke, line 237: can errors have lifetimes?

src/main.rs
```diff
@@ -141,35 +140,55 @@ async fn main() -> eyre::Result<()> {
...
             } else if api {
                 PackageType::Api
             } else {
                 PackageType::Impl
             };
 
-            command::init(r#type, package).await
+            command::init(kind, package)
+                .await
+                .wrap_err("init command failed")
```
mara-schulke, line 153: Please make them more meaningful eg. "Initialization of `<project-name>` failed

src/manifest.rs
```diff
@@ -78,30 +83,53 @@ pub struct Manifest {
...
-            .wrap_err("Failed to read manifest")?;
-
-        let raw: RawManifest = toml.parse().wrap_err("Failed to parse manifest")?;
+    pub async fn read() -> miette::Result<Self> {
+        Self::read_from(MANIFEST_FILE).await
+    }
 
-        Ok(raw.into())
+    /// Loads the manifest from the given path
+    pub async fn read_from(path: impl AsRef<Path>) -> miette::Result<Self> {
```
mara-schulke, line 99: can we merge them again? this isnt used anywhere right now, correct?

src/package.rs
```diff
@@ -48,35 +51,47 @@ impl PackageStore {
...
+    pub async fn create() -> miette::Result<()> {
         let create = |dir: &'static str| async move {
-            fs::create_dir_all(dir).await.wrap_err(eyre::eyre!(
-                "Failed to create dependency folder {}",
-                Path::new(dir).canonicalize()?.to_string_lossy()
-            ))
+            fs::create_dir_all(dir)
+                .await
+                .into_diagnostic()
+                .wrap_err_with(|| format!("failed to create {dir} directory"))
```
mara-schulke, line 59: miette!

src/package.rs
```diff
@@ -48,35 +51,47 @@ impl PackageStore {
...
-    pub async fn unpack(package: &Package) -> eyre::Result<()> {
+    pub async fn unpack(package: &Package) -> miette::Result<()> {
         let mut tar = Vec::new();
 
         let mut gz = flate2::read::GzDecoder::new(package.tgz.clone().reader());
 
         gz.read_to_end(&mut tar)
-            .wrap_err("Failed to decompress package")?;
+            .into_diagnostic()
+            .wrap_err_with(|| format!("failed to decompress package {}", package.name()))?;
```
mara-schulke, line 94: miette!

src/package.rs
```diff
@@ -103,63 +129,58 @@ impl PackageStore {
...
     /// Uninstalls a package from the local file system
-    pub async fn uninstall(package: &PackageName) -> eyre::Result<()> {
+    pub async fn uninstall(package: &PackageName) -> miette::Result<()> {
         let pkg_dir = Path::new(Self::PROTO_VENDOR_PATH).join(package.as_str());
 
         fs::remove_dir_all(&pkg_dir)
             .await
-            .wrap_err_with(|| format!("Failed to uninstall {package}"))
+            .into_diagnostic()
+            .wrap_err_with(|| format!("failed to uninstall package {package}"))
```
mara-schulke, line 142:
```suggestion
            .wrap_err(miette!("failed to uninstall package {package}"))
```

src/package.rs
```diff
@@ -170,7 +191,8 @@ impl PackageStore {
             let as_str: String = manifest
                 .clone()
                 .try_into()
-                .wrap_err("failed to render manifest as TOML")?;
+                .into_diagnostic()
+                .wrap_err_with(|| SerializationError(ManagedFile::Manifest))?;
```
mara-schulke, line 195:
```suggestion
                .wrap_err(SerializationError(ManagedFile::Manifest))?;
```

src/package.rs
```diff
@@ -179,53 +201,61 @@ impl PackageStore {
...
-                .wrap_err("Failed to pack tar")?,
+                .into_diagnostic()
+                .wrap_err("serialized manifest was too large to fit in a tarball")?,
         );
         header.set_mode(0o444);
         archive
             .append_data(&mut header, MANIFEST_FILE, Cursor::new(manifest_bytes))
-            .wrap_err("Failed to add manifest to release")?;
+            .into_diagnostic()
+            .wrap_err("failed to add manifest to release")?;
```
mara-schulke, line 211: miette!

src/package.rs
```diff
@@ -394,52 +433,50 @@ impl Default for PackageType {
...
-            "Invalid package name: {value} - only ASCII lowercase alphanumeric characters and dashes are accepted",
-        );
-        ensure!(
-            value
-                .get(0..1)
-                .wrap_err("Expected package name to be non empty")?
-                .chars()
-                .all(|c| c.is_ascii_alphabetic()),
-            "Package names must begin with an alphabetic letter"
+            value.chars().next().expect(UNEXPECTED_MSG).is_alphabetic(),
```
mara-schulke, line 454: Please dont expect

src/registry/artifactory.rs
```diff
@@ -174,35 +184,11 @@ impl Artifactory {
             package.version(),
         )
         .parse()
-        .wrap_err("Failed to construct artifact uri")?;
+        .expect("unexpected error: failed to construct artifact URL");
```
mara-schulke, line 187: please dont expect

src/package.rs
```diff
@@ -179,53 +201,61 @@ impl PackageStore {
...
             );
 
             archive
                 .append_data(
                     &mut header,
-                    entry.strip_prefix(&pkg_path).wrap_err_with(|| {
-                        format!("Failed to resolve path for entry {}", entry.display())
-                    })?,
+                    entry.strip_prefix(&pkg_path).expect(
+                        "unexpected error: collected file path is not under package prefix",
```
mara-schulke, line 233: Please dont expect

mara-schulke (changes requested):
Please dont use `expect` + cleanup wrap_err statements to be all `wrap_err(miette!())`

src/registry/artifactory.rs
asmello: Thanks for the detailed explanation, Mara! This makes it all much clearer.

From what I gather, there are two points where the code looks imperative:
* The signature of `make_auth_request()`, which takes an Option argument.
  * Here I agree a builder pattern would be nicer, though I'm a little skeptical about the complexity trade-off. Still, that's something I won't argue against.
* The construction of the request, which conditionally re-binds the builder variable a couple of times.

I believe the second part looks imperative is because the `reqwest` crate chose to implement `RequestBuilder` methods consume the builder, so we are forced to either chain only once or reassign. Ideally the builder API would take optionals so we could apply the functional style unconditionally whether the token and body are present or not.

I think moving the request building logic into a separate function is an ok change to make, although it is never used in isolation it does constitute a self-contained unit of behaviour.

Ultimately, IMO the nice thing here would be to define a ValidatedRequest struct that encapsulates the underlying `reqwest` request. Then we could fix both problems at once:

```rust
impl Artifactory {
    fn new_request(&self, method: Method, url: Url) -> ValidatedRequestBuilder {
        let mut builder = ValidatedRequest::new(method, url);
        if let Some(token) = &self.token {
            builder = builder.token(token);
        }
        builder
    }

    pub fn publish(&self, package: Package, repository: String) -> miette::Result<()> {
        // ...
        let response = self.new_request(method, url).body(data).send().await?;
        // ...
    }
}
```

A few notes:
* Functions don't need to specify the `body` argument if it's not used.
* Functions never need to specify the token argument explicitly, so there's no chance of forgetting to do that.
* `ValidatedRequest::send()` takes care of common response validation internally.
* We could make `new_request()` look more functional internally by making `.token()` take an `Option<String>`, but I think that's a bad trade-off.
  * We could also make `.token()` take a mutable reference to the builder (instead of consuming it) so that variable reassignment wouldn't be necessary. Not sure if that's worth the trade-off either, but I don't feel strongly about this point.

Honestly, a lot of the problems here are due to `reqwest`'s shortcomings. If we could provide the token on client construction (rather than at request construction), we wouldn't need the `new_request()` method. If we could define common validation logic, we wouldn't need the custom `ValidatedRequest` type either.

I'll go ahead with this style, and happy to adjust as necessary, but please let's not focus too much on this as improving this code is really not the goal of this PR.

src/lock.rs
asmello: Not sure what you mean, errors are just normal types, so I guess yes.

src/manifest.rs
asmello: It is, actually, in `PackageStore::resolve()`.

src/main.rs
asmello: This is actually impossible without either a refactor that pulls the project name resolution logic out of `command::init()` and into `main()` or by returning a typed error from the command. I don't think either is worth the trouble so marking as resolved.

src/registry/artifactory.rs
asmello: Ok, we chatted a bit offline and with some additions (and a less confusing name for the new request type) we agreed this is best.

src/lib.rs
```diff
@@ -30,12 +33,14 @@ pub mod package;
 /// Supported registries
 pub mod registry;
 
+mod managed_file;
```
mara-schulke, line 36: Can rename this to smth more generic? it doesnt seem sensible to have a whole mod for _one_ enum. Maybe this can be just toplevel in the lib and be pub(crate)?

src/main.rs
```diff
@@ -141,35 +140,55 @@ async fn main() -> eyre::Result<()> {
...
             registry,
             dependency,
-        } => command::add(registry, &dependency).await,
-        Command::Remove { package } => command::remove(package).await,
+        } => command::add(registry, &dependency)
+            .await
+            .wrap_err(miette!("add command failed")),
+        Command::Remove { package } => command::remove(package)
+            .await
+            .wrap_err(miette!("remove command failed")),
```
mara-schulke, line 163: these are still bad error messages. It would be neat to do something meaningful here like `failed to remove `{package}` from `Proto.toml`` - smth like this should be the toplevel message for all of the commands

src/managed_file.rs
```diff
@@ -0,0 +1,26 @@
+use std::fmt::Display;
+
+use crate::{credentials::CREDENTIALS_FILE, lock::LOCKFILE, manifest::MANIFEST_FILE};
+
+#[derive(Debug)]
+pub(crate) enum ManagedFile {
```
mara-schulke, line 6: im just put this in the lib.rs for now

src/manifest.rs
```diff
@@ -12,21 +12,26 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use eyre::Context;
+use miette::IntoDiagnostic;
+use miette::{miette, Context};
```
mara-schulke, line 16: can be combined imports

src/manifest.rs
mara-schulke: Ah! Oka, well, valid reasoning :)

src/package.rs
```diff
@@ -48,35 +51,47 @@ impl PackageStore {
...
+    pub async fn create() -> miette::Result<()> {
         let create = |dir: &'static str| async move {
-            fs::create_dir_all(dir).await.wrap_err(eyre::eyre!(
-                "Failed to create dependency folder {}",
-                Path::new(dir).canonicalize()?.to_string_lossy()
-            ))
+            fs::create_dir_all(dir)
+                .await
+                .into_diagnostic()
+                .wrap_err_with(|| miette!("failed to create {dir} directory"))
```
mara-schulke, line 59:
```suggestion
                .wrap_err(miette!("failed to create {dir} directory"))
```

src/package.rs
```diff
@@ -48,35 +51,47 @@ impl PackageStore {
...
-    pub async fn unpack(package: &Package) -> eyre::Result<()> {
+    pub async fn unpack(package: &Package) -> miette::Result<()> {
         let mut tar = Vec::new();
 
         let mut gz = flate2::read::GzDecoder::new(package.tgz.clone().reader());
 
         gz.read_to_end(&mut tar)
-            .wrap_err("Failed to decompress package")?;
+            .into_diagnostic()
+            .wrap_err_with(|| miette!("failed to decompress package {}", package.name()))?;
```
mara-schulke, line 94: i would use the short syntax everywhere, we dont have performance problems at all, imo readability is more important

src/package.rs
```diff
@@ -87,12 +102,23 @@ impl PackageStore {
...
+                )
+            })?;
 
-        tar.unpack(pkg_dir.clone()).wrap_err(format!(
-            "Failed to unpack tar of {}",
-            package.manifest.package.name
-        ))?;
+        tar.unpack(pkg_dir.clone())
+            .into_diagnostic()
+            .wrap_err_with(|| {
```
mara-schulke, line 115: so talking abt all of these

src/package.rs
```diff
@@ -331,17 +371,29 @@ impl TryFrom<Bytes> for Package {
...
-        let manifest = manifest.bytes().collect::<io::Result<Vec<_>>>()?;
-        let manifest = String::from_utf8(manifest)?
+            .ok_or_else(|| miette!("missing manifest"))?;
+
+        let manifest = manifest
+            .bytes()
+            .collect::<io::Result<Vec<_>>>()
+            .into_diagnostic()
+            .wrap_err_with(|| DeserializationError(ManagedFile::Manifest))?;
+        let manifest = String::from_utf8(manifest)
```
mara-schulke, line 381:
```suggestion

        let manifest = String::from_utf8(manifest)
```

src/package.rs
```diff
@@ -331,17 +371,29 @@ impl TryFrom<Bytes> for Package {
...
-
-        let manifest = manifest.bytes().collect::<io::Result<Vec<_>>>()?;
-        let manifest = String::from_utf8(manifest)?
+            .ok_or_else(|| miette!("missing manifest"))?;
+
+        let manifest = manifest
+            .bytes()
+            .collect::<io::Result<Vec<_>>>()
+            .into_diagnostic()
+            .wrap_err_with(|| DeserializationError(ManagedFile::Manifest))?;
```
mara-schulke, line 380:
```suggestion
            .wrap_err(DeserializationError(ManagedFile::Manifest))?;
```

src/registry/artifactory.rs
```diff
@@ -26,191 +27,166 @@ pub struct Artifactory {
...
-        }
-
-        let response = request.send().await?;
-
-        let status = response.status();
+        self.new_request(Method::GET, repositories_url)
+            .send()
+            .await
+            .map(|_| ())
+            .map_err(miette::Report::from)
```
mara-schulke, line 65: neat!

src/registry/artifactory.rs
```diff
@@ -26,191 +27,166 @@ pub struct Artifactory {
...
+        let artifact_url = resolve_url(dependency.clone())?;
 
-        ensure!(status.is_success(), "Failed to ping artifactory");
+        tracing::debug!("Hitting download URL: {artifact_url}");
 
-        tracing::debug!("Pinging artifactory succeeded");
+        let response = self.new_request(Method::GET, artifact_url).send().await?;
 
-        Ok(())
+        convert_to_package(response)
```
mara-schulke, line 76: Hm, i have to say i dislike this, also the logic isnt reused right? Cant we just inline the logic for creating a pkg from the request?

src/registry/artifactory.rs
```diff
@@ -26,191 +27,166 @@ pub struct Artifactory {
...
-            let mut url = dependency.manifest.registry.clone();
-            url.set_path(&format!(
-                "{}/{}/{}/{}-{}.tgz",
-                path,
-                dependency.manifest.repository,
-                dependency.package,
-                dependency.package,
-                version
-            ));
+struct RequestBuilder(reqwest::RequestBuilder);
```
mara-schulke, line 114: why do we have this, this doesnt add any logic right?

src/registry/artifactory.rs
```diff
@@ -26,191 +27,166 @@ pub struct Artifactory {
...
+    fn body(mut self, payload: impl Into<Body>) -> Self {
+        self.0 = self.0.body(payload);
+        self
+    }
 
-        if let Some(token) = &self.token {
-            request = request.bearer_auth(token);
-        }
+    async fn send(self) -> miette::Result<ValidatedResponse> {
+        self.0.send().await.into_diagnostic()?.try_into()
```
mara-schulke, line 132: Ah oka, i get it, but tbh, i dont see too much value in this

src/registry/artifactory.rs
```diff
@@ -26,191 +27,166 @@ pub struct Artifactory {
...
+    }
+}
 
-        let response = request.send().await?;
+struct ValidatedResponse(reqwest::Response);
```
mara-schulke, line 136: i would impl also `into_inner`, its common for wrappers

src/registry/artifactory.rs
```diff
@@ -26,191 +27,166 @@ pub struct Artifactory {
...
-            .await
-            .wrap_err("Failed to upload release to artifactory")?;
+    Ok(url.into())
+}
 
-        ensure!(
-            !response.status().is_redirection(),
-            "Remote server attempted to redirect publish request - is the registry URL valid?"
-        );
+async fn convert_to_package(response: ValidatedResponse) -> miette::Result<Package> {
```
mara-schulke, line 176: i dislike the fundamental changes to the artifactory impl, this looks way more confusing to me as it did before

src/registry/local.rs
```diff
@@ -33,37 +39,9 @@ impl LocalRegistry {
         LocalRegistry { base_dir }
     }
 
-    pub async fn download(&self, dependency: Dependency) -> eyre::Result<Package> {
+    pub async fn download(&self, dependency: Dependency) -> miette::Result<Package> {
```
mara-schulke, line 42: doc comments

src/registry/local.rs
```diff
@@ -33,37 +39,9 @@ impl LocalRegistry {
...
         // TODO(rfink): Factor out checks so that artifactory and local registry both use them
```
mara-schulke, line 43: remove the todo?

mara-schulke (changes requested):
I dislike the fact that this mr does things way beyond its scope, its a bit hard to review.

The error side looks good to me mostly! Just the refactoring of existing code makes changes hard to track and bloats this. Could you eventually split refactoring the registries etc in a second pr, based on this one and revert the changes for now so we can merge this today?

Mostly lgtm! 😊

src/lib.rs
asmello: I think it'd be best to make a module of this and move the `manifest.rs`, `credentials.rs` and `lock.rs` modules there. Happy to move this to the lib mod for now but does that seem sensible?

src/main.rs
asmello: I replied to your earlier comment - this is actually very hard with the current architecture, we'd need to either return a structured error from the command or move package name resolution logic to `main.rs`. I prefer to not attempt that in this PR, it's grown twice as large as originally intended already.

src/manifest.rs
asmello: Weird, probably an artefact of a recent merge.

src/lib.rs
mara-schulke: Yes, i agree! Once we merge them, lets put it there

src/package.rs
asmello: I think you actually suggested some of these, but I can revert.

src/main.rs
mara-schulke: Huh? How is it hard?

You can just change the message:

```suggestion
            .wrap_err(miette!("failed to remove `{package}` from `{MANIFEST_FILE}`")),
```

src/registry/artifactory.rs
asmello: This was a poor man's attempt at minimising the risk someone would make a package out of a non-validated response, by encapsulating that logic into a standalone function, but can revert this back to inline code.

src/registry/artifactory.rs
asmello: In what?

src/registry/artifactory.rs
asmello: I impl `into()` instead, isn't that sufficient?

src/registry/artifactory.rs
asmello: We discussed this, this is the Request wrapper that makes it so we can't make requests without validating the response.

src/registry/artifactory.rs
asmello: Two reasons for this refactor:
1. Splits out a self-contains unit of behaviour, making the download implementation much easier to understand.
2. Makes it less likely people will end up re-implementing the conversion from response to package and use a non-validated response.
Happy to revert to inline, but this was inspired by your suggestion to prevent people from using raw bytes to create Packages.

src/main.rs
asmello: For this one command, yes, we know the package name because it was passed as an argument, but for many others that is not available at this level. So this is hard:

> smth like this should be the toplevel message for all of the commands

src/registry/artifactory.rs
asmello: I can replace the `From<ValidatedResponse> for reqwest::Response` implementation if you want, but it's pretty much either/or, otherwise we end up with dead code.

mara-schulke (approved):

src/registry/artifactory.rs
asmello: #103

src/main.rs
asmello: #129

Issue #108, opened 2023-10-10
Credentials storage does not respect the XDG specification


xfbs:
Currently, the credentials are stored in a hard-coded path determined by the `$BUFFRS_HOME` variable, or in `~/.buffrs/credentials.toml`, see [this code](https://github.com/helsing-ai/buffrs/blob/main/src/credentials.rs#L35C9-L35C9). While this is initially a sensible approach, there are some improvements that could be made.

## Issues

Depending on the operating system that you are using, you may not want to store things in your home directory. Some operating systems do not understand the concept of dotfiles (files or folders, which are hidden by default due to a leading dot in their name). Some people may prefer to not clobber their home directories with configuration, but rather store it somewhere else. 

In general, it is warmly recommended to adhere to a standard when it comes to storing configuration rather than "clobbering" user's home directories with it. See also

## Standard

For this reason, the XDG group has a specification called the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) which sets standards that applications should respect. One of these standards defines where configuration should be stored. The default location as per this standard would be `~/.config/buffrs/credentials.toml`, but the spec also defines environment variables that can be set to change this.

Similarly, Windows has the [KNOWNFOLDERID](https://learn.microsoft.com/en-us/windows/win32/shell/knownfolderid?redirectedfrom=MSDN) API and macOS the [Standard Directories](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html#//apple_ref/doc/uid/TP40010672-CH2-SW6).

## Solution

We can use the [dirs](https://github.com/dirs-dev/dirs-rs) crate to consume these standards. Another crate offered by the same team is [directories](https://github.com/dirs-dev/directories-rs), which should make it even easier for us.

## Notes

Note that many applications still do not respect the XDG standard, including my beloved editor `vim`. It is not the end of the world if we do not adhere to them. But I do believe that it is a good standard and that there are few reasons not to stick to it. Ideally, we can stick to it in a backwards-compatible way, such that the application would prefer configuration created in the XDG-compatible paths, but would also search the current locations.

## See Also

- [Defending `$HOME`](https://soc.me/standards/defending-home.html)
- [`$HOME`, not so sweet `$HOME`](https://gist.github.com/sharadhr/39b804236c1941e9c30d90af828ad41e#213xdg-base-directories)

[...]

PR #109, opened 2023-10-10
Credentials storage XDG

xfbs:
See #108.

This PR switches the code used for reading and writing credentials to being XDG-compatible, but also backwards-compatible.

It does that by having a list of candidate paths for the configuration that it checks, preferring the XDG-compatible path but falling back to the previous path. When writing, it will only write to the XDG-compatible path.

This PR is still a draft.

### DIFF
--- a/src/credentials.rs
+++ b/src/credentials.rs
@@ -14,13 +14,19 @@
 
 use eyre::{Context, ContextCompat};
 use serde::{Deserialize, Serialize};
-use std::{collections::HashMap, env, path::PathBuf};
+use std::{
+    collections::HashMap,
+    env::{var, VarError},
+    io::ErrorKind,
+    path::PathBuf,
+};
 use tokio::fs;
 
 use crate::registry::RegistryUri;
 
 /// Global configuration directory for `buffrs`
 pub const BUFFRS_HOME: &str = ".buffrs";
+pub const BUFFRS_DIR: &str = "buffrs";
 /// Filename of the credential store
 pub const CREDENTIALS_FILE: &str = "credentials.toml";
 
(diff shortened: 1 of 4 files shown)


asmello:
Overall comment: while resolving the appropriate config directory in a portable way is a great idea, we currently don't have support for actual configuration in Buffrs (see #94 for a discussion on how that might look like). The credentials file is technically data, not configuration, so I think `dirs::data_dir()` would be a more appropriate choice. Though I believe the current approach [mimics Cargo](https://doc.rust-lang.org/cargo/guide/cargo-home.html), so it'd be interesting to do some research into why they went that way rather than use the OS defaults.

From another side, I actually think that we should reconsider using keychains for credential management (this feature was removed in #8). Even though stable Cargo uses a plaintext file, nightly has already implemented keychains as the default credential storage (https://github.com/rust-lang/cargo/issues/8933), so there must be a way to make this work sensibly.

mara-schulke:
@asmello i think, this could be sensible, as the host machine should be ultimately seen as secure to store a token. Thus I would not prioritize it. Maybe create a discussion?

asmello:
> But the host machine should be ultimately seen as secure to store a token

I disagree with this statement, any sensitive data should be encrypted at rest, that's just regular AppSec. Any program that does otherwise has weakness [CWE-256](https://cwe.mitre.org/data/definitions/256.html) (or [CWE-312](https://cwe.mitre.org/data/definitions/312.html)) and cannot be considered secure by any official standard. But we don't need to block Patrick's contribution based on that (I just figured I'd point out that the file approach may be superseded in the future).

Created #119

xfbs:
> The credentials file is technically data, not configuration, so I think `dirs::data_dir()` would be a more appropriate choice.

I don't know if I would agree with that! In my understanding, credentials is something that is explicitly configured by the user, and thus not data. In my understanding, data would be more static things that are needed at runtime. Anything that you need to back up when you migrate to a new machine and want to have your tools work should be configuration.

> Though I believe the current approach [mimics Cargo](https://doc.rust-lang.org/cargo/guide/cargo-home.html), so it'd be interesting to do some research into why they went that way rather than use the OS defaults.

There is a [tracking issue](https://github.com/rust-lang/cargo/issues/1734) on this. I think that the tl;dr is: XDG standard adoption is spotty, most UNIXy tools default to just putting things in dotfiles, and this works. I  think that for new tools, the XDG paths should be respected, just like for new web development it should use ARIA labels to be accessible, but one should expect to see a lot of projects that do not prioritize it.

> I actually think that we should reconsider using keychains for credential management.

This is not a terrible idea. Although, maybe it should be done in a separate step. To have it where we check:

1. System-local keychain (macOS keychain, GNOME keychain, whatever Windows does)
2. XDG-path (`$XDG_CONFIG_DIR/buffrs/credentials.toml`)
3. Dotfile location (`~/.buffrs/credentials.toml`)
4. System-wide config (`/etc/buffrs/credentials.toml`)

And then you can basically use whatever you want, but we obviously prefer the keychain approach for saving new credentials, if your system happens to have a standardized keychain store. 

It might be a nightmare to support it however, if there is not a crate with a nice abstraction.

src/credentials.rs
```diff
@@ -32,54 +38,78 @@ pub struct Credentials {
...
+    /// Get buffrs config directory.
+    fn buffrs_home() -> Result<PathBuf, VarError> {
+        var("BUFFRS_HOME").map(PathBuf::from)
+    }
 
-        let home = env::var("BUFFRS_HOME").map(PathBuf::from).unwrap_or(home);
+    /// XDG path: uses whichever config directory is appropriate for the platform.
+    ///
+    /// For example, on Linux this might be `~/.config/buffrs/credentials.toml`.
+    fn xdg_path() -> PathBuf {
```
asmello, line 49: I'd call this "config_path" or something similar. Knowledge of the XDG spec shouldn't be required to understand what this is about.

src/credentials.rs
```diff
@@ -32,54 +38,78 @@ pub struct Credentials {
...
+    fn possible_locations() -> Vec<PathBuf> {
+        vec![Self::xdg_path(), Self::legacy_path()]
     }
 
     /// Checks if the credentials exists
     pub async fn exists() -> eyre::Result<bool> {
-        fs::try_exists(Self::location()?)
-            .await
-            .wrap_err("Failed to detect credentials")
+        for location in &Self::possible_locations() {
```
asmello, line 69: I think this may be cleaner in functional style: `Self::possible_locations().iter().map(...).all()`.

src/credentials.rs
```diff
@@ -32,54 +38,78 @@ pub struct Credentials {
...
+            if fs::try_exists(&location).await? {
```
asmello, line 70: Do we really want to terminate on the first error? Imagine we fail to access the system-wide credentials because we lack permission, shouldn't we keep going and try user locations too?

BTW `fs::try_exists` [is only in nightly](https://doc.rust-lang.org/stable/std/fs/fn.try_exists.html) so I'm surprised this compiles in CI (probably something worth looking into, will file an issue for it).

src/credentials.rs
```diff
@@ -32,54 +38,78 @@ pub struct Credentials {
...
+                return Ok(true);
+            }
+        }
+
+        return Ok(false);
```
asmello, line 75: Flagged by clippy, this return is superfluous.

asmello (commented):
Didn't comment on error messages as the way Buffrs handles that is due to change.

src/credentials.rs
asmello: Womp, you're right! This is why I don't like tokio shadowing std stuff... 😅  Ignore that part (and I'll close the issue).

xfbs:
Something you can look into @asmello: https://docs.rs/keyring/latest/keyring/. This crate purports to be a cross-platform keychain API, might do just what we need.

> My main concern with using the config path, other than semantics, is security — this directory is often backed up and synced across devices, and implicitly sharing credentials this way is potentially not intended.

There are people doing strange stuff, but I think that generally, anything under `$HOME` is security sensitive. I have lots of secrets in `~/.config`, `~/.ssh`, `~/.mozilla`, `~/.gpg`, some of which are unencrypted (LUKS ftw). If people back this up and store it unencrypted or sync it willy-nilly, then that is potentially an error on their part rather than on ours.

The only scheme of syncing dotfiles that I know (and practise) is putting them into a git repo and cloning it on new machines. This is explicitly a one-way sync kind of thing. You have to explicitly add, commit and push things you want to sync, and never all of `~` or all of `~/.config`. That would be a terrible idea, regardless of where we put our credentials!

asmello:
> There are people doing strange stuff, but I think that generally, anything under `$HOME` is security sensitive. I have lots of secrets in `~/.config`, `~/.ssh`, `~/.mozilla`, `~/.gpg`, some of which are unencrypted (LUKS ftw). If people back this up and store it unencrypted or sync it willy-nilly, then that is potentially an error on their part rather than on ours.

Let's make a distinction between dot-dirs like `~/.ssh` and `~/.gpg`, the entire `$HOME` directory and the XDG config path. We're discussing the last one, which by definition shouldn't contain any sensitive information. Arguably the data directory shouldn't either, because this wasn't explicitly intended as per the spec, although one can reasonably extend the definition of sensitive to cover all user data (as has been the trend with GDPR and other regulatory policies).

So to be clear:
* Yes, we can consider `$HOME`, `~/.ssh`, `~/.mozilla`, `~/.gpg` (and `~/.buffrs`) all to be sensitive and expect users not to sync them across devices they deem insecure.
* The same cannot be said about `~/.config`, as it's supposed to only have configuration data.
  * Although I agree a cautious user will still avoid syncing this directory since there's nothing enforcing that apps follow best practices.

BTW this is not just my opinion, the issues I linked (especially the AWS one) show that this is a common interpretation.

xfbs:
Closing due to offline discussion, long-term `buffrs` will need to add a cache anyways and likely require a `~/.buffrs` folder.

PR #124, opened 2023-10-12
Use serde_typename crate for DigestAlgorithm

xfbs:
This uses the `strum` crate for providing the `FromStr` and `Display` implementations of DigestAlgorithm.

This leads to more maintainable code, because there is only one place where the representation is defined (as an annotation), and we do not have to implement traits by hand.

There is one alternative approach, which might also be useful, so I will describe it here: instead of using `strum`, we could also use `serde`. We do this by deriving `Serialize` and `Deserialize` on the enum, and then using `serde_typename` for the implementation. For example:

```rust
impl FromStr for DigestAlgorithm {
    type Err = serde_typename::ParseError;

    fn from_str(input: &str) -> Result<Self, Self::Err> {
        serde_typename::from_str(input)
    }
}
```

The advantage of this is:
- Not needing to pull in another crate
- FromStr/Display implementation is synced with the serde serialization representation

The downsides of this are:
- More code (still need to manually implement FromStr and Display)

### DIFF
--- a/src/lock.rs
+++ b/src/lock.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::{collections::HashMap, str::FromStr};
+use std::{collections::HashMap, fmt, str::FromStr};
 
 use miette::{ensure, miette, Context, IntoDiagnostic};
 use ring::digest;
@@ -30,22 +30,13 @@ use crate::{
 
 /// File name of the lockfile
 pub const LOCKFILE: &str = "Proto.lock";
-const SHA256_TAG: &str = "sha256";
 
 /// Supported types of digest algorithms
-#[derive(Clone, PartialEq, Eq)]
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub enum DigestAlgorithm {
     /// 256 bits variant of SHA-2
+    #[serde(rename = "sha256")]
     SHA256,
-}
-
-impl std::fmt::Display for DigestAlgorithm {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        let s = match self {
-            Self::SHA256 => SHA256_TAG,
-        };
-        f.write_str(s)
-    }
 }
 
 /// Represents a ring digest algorithm that isn't supported by Buffrs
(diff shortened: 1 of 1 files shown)


asmello:
This does clear things up nicely, but unsure if it's worth adding a whole new dependency. The DigestAlgorithm enum is not likely to change very often, I expect the manual implementations will require very little overhead. My bias is to keep things as they are, but I defer to @mara-schulke's judgement here.

xfbs:
If adding a dependency is the issue, then I would go with the `serde_typename` route. That crate is already a dependency and is used elsewhere and it still brings a benefit (maintenance, and keeping it DRY).

mara-schulke:
Imo let's use serde_typename, its much more lightweight than strum

xfbs:
I've changed this PR to use `serde_typename` instead.

src/lock.rs
```diff
@@ -27,30 +27,25 @@ use crate::{
...
-
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
-        match s {
-            SHA256_TAG => Ok(Self::SHA256),
-            other => eyre::bail!("invalid digest algorithm: {other}"),
-        }
+impl fmt::Display for DigestAlgorithm {
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // this will not fail as long as `DigestAlgorithm` is an enum,
+        let typename = serde_typename::to_str(self).expect("digest algorithm type name");
```
mara-schulke, line 47: Can we map this as a format result error and not expect?

xfbs: We technically can, but [std::fmt::Error](https://doc.rust-lang.org/stable/std/fmt/struct.Error.html) does not let us add any information about what went wrong:

> This type does not support transmission of an error other than that an error occurred. Any extra information must be arranged to be transmitted through some other means.

Realistically, this can never panic, as long as `DigestAlgorithm` is an enum. We can check that in unit tests. Perhaps a semantically better way to write this would be:

```rust
match serde_typename::to_string(self) {
    Ok(value) => value,
    Err(error) => unreachable!("DigestAlgorithm failed to convert to string: {error}"),
}
```

That way we would convey that this is not something that should ever occur.

asmello: For what it's worth I think it's perfectly reasonable to expect here, it's what expect is intended for. But I've had this discussion with Mara before.

xfbs: I have changed it to use `unreachable!()`, I find that that is more clear anyways. Whenever I see `expect`, it makes me wonder if someone forgot to do error handling here or if this is truly something that cannot happen. So `unreachable` is a lot more clear in this case.

xfbs:
Just rebased this and cleaned up a few bits.

src/lock.rs
```diff
@@ -56,10 +47,19 @@ pub struct UnsupportedAlgorithm(String);
...
+            _other => Err(UnsupportedAlgorithm(input.into())),
+        }
+    }
+}
+
+impl fmt::Display for DigestAlgorithm {
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match serde_typename::to_str(self) {
+            Ok(name) => fmt.write_str(name),
+            Err(error) => unreachable!("cannot convert DigestAlgorithm to string: {error}"),
```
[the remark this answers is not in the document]
mara-schulke: https://doc.rust-lang.org/std/fmt/struct.Error.html

ah i see it is useless

asmello (approved):
I don't think this is necessarily an improvement but don't see any harm either.

PR #126, opened 2023-10-12
Allows building registry

xfbs:
This allows for compiling the registry.

- Adds `tonic` and `prost` dependencies. These are needed for the `buffrs` generated code to compile correctly.
- Adds `registry` to crate workspace. This allows for a better dev experience (they share a `target` folder, meaning that you don't have to recompile shared dependencies twice) and allows the tests to be included in the CI.

Some open questions:

- I've had to add `tonic` and `prost` as direct dependencies of the registry for it to build. Is that the way to go? We could circumvent this, by re-exporting `tonic` and `prost` from buffrs (as `buffrs::tonic` and `buffrs::prost`) and minimally rewriting generated files. This might allow for a smoother experience, because it means you are not confronted with errors.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,3 +66,6 @@ predicates = "3.0"
 pretty_assertions = "1.4"
 ring = "0.16.20"
 hex = "0.4.3"
+
+[workspace]
+members = [".", "registry"]
--- a/registry/src/proto.rs
+++ b/registry/src/proto.rs
@@ -1,3 +1,3 @@
 // (c) Copyright 2023 Helsing GmbH. All rights reserved.
 
-buffrs::include!();
+::buffrs::include!();
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -122,6 +122,28 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-stream"
+version = "0.3.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cd56dd203fef61ac097dd65721a419ddccb106b2d2b70ba60a6b529f03961a51"
+dependencies = [
+ "async-stream-impl",
+ "futures-core",
+ "pin-project-lite",
+]
+
+[[package]]
+name = "async-stream-impl"
+version = "0.3.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "16e62a023e7c117e27523144c5d2459f4397fcc3cab0085af8e2224f643a0193"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "async-trait"
 version = "0.1.73"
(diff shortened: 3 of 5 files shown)


Cargo.toml
```diff
@@ -66,3 +66,6 @@ predicates = "3.0"
 pretty_assertions = "1.4"
 ring = "0.16.20"
 hex = "0.4.3"
+
+[workspace]
+members = [".", "registry"]
```
mara-schulke, line 71: does this affect the publishing of `buffrs`?

registry/src/proto.rs
```diff
@@ -1,3 +1,3 @@
 // (c) Copyright 2023 Helsing GmbH. All rights reserved.
 
-buffrs::include!();
+::buffrs::include!();
```
mara-schulke, line 3: why is this needed?

Cargo.toml
xfbs: It does not — when you do `cargo publish`, it always only publishes the crate you are currently in. You can verify this by doing a `cargo publish --dry-run`.

When we add a dependency from `registry` to `buffrs`, we can do it like this:

```toml
buffrs = { path = "..", version = "0.1.0" }
```

What this does is that if you are building locally, it'll just use whatever the path points to (it validates that it is the right version). But if you are publishing the registry (should we want to do that), then it strips out the `path` and goes only by the `version`. That way it's easy to work locally but when we publish it does what it should.

registry/src/proto.rs
xfbs: I had an error if I did not include that:

```
error[E0433]: failed to resolve: could not find `include` in `buffrs`
 --> registry/src/proto.rs:3:9
  |
3 | buffrs::include!();
  |         ^^^^^^^ could not find `include` in `buffrs`

error[E0659]: `buffrs` is ambiguous
  --> registry/src/proto.rs:3:1
   |
3  | buffrs::include!();
   | ^^^^^^ ambiguous name
   |
   = note: ambiguous because of a conflict between a macro-expanded name and a less macro-expanded name from outer scope during import or macro resolution
   = note: `buffrs` could refer to a crate passed with `--extern`
   = help: use `::buffrs` to refer to this crate unambiguously
note: `buffrs` could also refer to the module defined here
  --> /Users/patrick.elsen/Projects/helsing/buffrs/target/debug/build/buffrs-registry-b5cc01ff6df2be9c/out/buffrs.rs:1:1
   |
1  | / pub mod buffrs {
2  | |     pub mod dependency {
3  | |         include!("buffrs.dependency.rs");
4  | |     }
...  |
16 | |     }
17 | | }
   | |_^
   = help: use `self::buffrs` to refer to this module unambiguously
```

I think the issue is that there are two `buffrs`: the crate `buffrs` and the module `buffrs` which is included here.

Perhaps we could rename the generated module to something like `buffers_gen` instead.

mara-schulke (approved):

PR #127, opened 2023-10-12
Hooks up database to registry

xfbs:
This hooks up the postgres database to the registry. Also provides a `docker-compose` file to launch the database locally. Currently, it should be possible to run `cargo run`, but it won't do anything. 

Depends on #126.


registry/src/config.rs
```diff
@@ -1 +1,14 @@
...
+use std::net::SocketAddr;
+
+#[derive(Parser, Clone, Debug)]
+pub struct Config {
+    /// Address to listen to for incoming connections.
+    #[clap(long, short, env, default_value = "0.0.0.0:8000")]
+    pub listen: SocketAddr,
+
+    /// URL of Postgres database to connect to.
+    #[clap(long, short, env, default_value = "postgres://buffrs:buffrs@localhost")]
```
mara-schulke, line 12: no default please 👀

registry/src/config.rs
```diff
@@ -1 +1,14 @@
...
+    pub database: String,
```
mara-schulke, line 13: `url::Url`

registry/src/db.rs
```diff
@@ -0,0 +1,12 @@
+// (c) Copyright 2023 Helsing GmbH. All rights reserved.
+use sqlx::PgPool;
```
mara-schulke, line 2:
```suggestion

use sqlx::PgPool;
```

registry/src/db.rs
```diff
@@ -0,0 +1,12 @@
+// (c) Copyright 2023 Helsing GmbH. All rights reserved.
```
mara-schulke, line 1: should be apache 2 (see the cli)

registry/src/config.rs
```diff
@@ -1 +1,14 @@
 // (c) Copyright 2023 Helsing GmbH. All rights reserved.
+use clap::Parser;
+use std::net::SocketAddr;
+
+#[derive(Parser, Clone, Debug)]
+pub struct Config {
```
mara-schulke, line 6: These should go in main imo, as they are binary related, the config.rs is usually for file based configs (at helsing)

registry/src/config.rs
xfbs: Because of the risk of forgetting to set it in production?

registry/src/db.rs
xfbs: I did not add those comments — they existed in the empty files. I am assuming you want me to copy the license headers from the `cli` and apply them to all of the source files under `registry`.

registry/src/config.rs
xfbs: I'll put this into an `options.rs` instead then. Don't like having it in the `main.rs` for abstraction reasons and because I don't like importing a lot of types in `main.rs`.

xfbs:
Just rebased on top of current `main`, and also:

- Copied the license header from the `buffrs` CLI into the registry.
- Adjusted spacing under the license header.
- Imported the `Url` type for the options.
- Renamed `config.rs` to `options.rs`.

registry/src/config.rs
mara-schulke: More because this is a weird default, when not developing (and its shown in the cli)

registry/src/db.rs
mara-schulke: Ah sorry, i did it then 😄

mara-schulke: Yes!

mara-schulke (approved):
Except for the defaults that only apply to development, this looks good!

registry/src/config.rs
xfbs: Not quite sure what you mean by a weird default. I think it is quite useful to have a default value here.

What I like to be able to do is launch it locally without needing a lot of configuration. For example, this is how I can launch it locally at the moment:

```rust
docker compose up # launch database
cargo run         # launch registry
```

Removing this default means that I need to either memorize the default, or introduce some kind of runner:

```rust
docker compose up
# this is quite annoying to have to type out every time
cargo run -- --listen postgres://buffrs:buffrs@localhost
```

There is one change we could make that might work here, which is to keep it but enable it only for development builds, by specifying it like this:

```rut
#[cfg_attr(dev, clap(default_value = "postgres://buffrs:buffrs@localhost"))]
```

The advantage of this is that it means there is no default value for production. I have had an instance before where an environment variable was spelled wrong, resulting in a configuration value being set incorrectly and that was quite annoying to debug. So in doing it this way, we preserve the ability to launch it locally easily while not having a default set in the production release.

Would this idea address your concern but still allow a good developer experience?

asmello:
I don't have anything to add, as long as prod mode doesn't use unsafe defaults I'm happy with this change.

registry/src/config.rs
mara-schulke: Yes, thank you!

registry/src/options.rs
```diff
@@ -0,0 +1,28 @@
...
+// limitations under the License.
+
+use clap::Parser;
+use std::net::SocketAddr;
+use url::Url;
+
+#[derive(Parser, Clone, Debug)]
+pub struct Options {
+    /// Address to listen to for incoming connections.
+    #[clap(long, short, env, default_value = "0.0.0.0:8000")]
```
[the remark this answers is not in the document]
mara-schulke: This would be neat, because we could have a standard port (not 8000) which we expect registries to run, allowing people to run a local registry proxy which does not interfere with their development setup (3000, 8000) or alike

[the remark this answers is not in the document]
xfbs:
```
$ passgen "[3-9][0-9]{3}"
4367
```

How about this one?

PR #132, opened 2023-10-18
Refactors `src/package.rs`

xfbs:
This PR refactors the module `package` in the buffrs CLI. Since I am looking to integrate the `validation` crate with `buffrs`, I saw some low-hanging fruits for code improvement. On a high level, the improvements introduced by this PR address the following:

- *async correctness*: not using blocking code in the path of async code. This is quite easy to get wrong in Rust, as of yet there is not much compiler help to avoid this (to my knowledge). In certain cases, it can be okay to mix the two (using a blocking mutex in a hot loop, writing a CLI that is single-threaded). But to build robust, reusable code we should try to get this right.
- *error handling*: error handling is something that can always use some love. This PR adds some error types in strategic places, mostly to make testing easier.
- *test coverage*: this PR attempts to increase net code coverage to 80%.
- *encapsulation*: trying to not leak implementation details when it is not necessary to leak them. This allows us freedom to change the implementation without breaking too many things.
- *robustness*: trying to avoid hidden assumptions by making them explicit. One example for this is the reliance on the current working directory in some of the code paths. 

I try to keep the changes in here to be mostly non-opinionated, so that they are uncontroversial. 

**Edit: this PR is now ready**. Here is a list of the major changes I have made here:

- [Consolidated the `impl PackageStore` blocks](https://github.com/helsing-ai/buffrs/pull/132/commits/3f387ead357b6a57a9bc7d0e09bc87618498aafd). There were multiple of them, I just squashed them together into a single one.
- [Removes superfluous `TryFrom` implementations](https://github.com/helsing-ai/buffrs/pull/132/commits/3e9a3b94ee736934ac2938352308a1aa9bdab624). `TryFrom` and `FromStr` are handy, but we should only implement the ones we need. For cases where a string slice suffices, `FromStr` is the way to go. For cases where ownership is needed, `TryFrom<String>` is the way to go. Always prefer `TryFrom<&str>` over `TryFrom<&String>`, because the former is more generic (the latter is just one specific container type for a `&str`), but usually neither are useful because what you really want is `FromStr`. Here I just removed some implementations that we are not using and that do not add value.
- [Implements custom error type for `PackageName` parsing](https://github.com/helsing-ai/buffrs/pull/132/commits/4e0e377efdefbf87146a8ce728bb66d48a18834e). Here I implemented a custom error type, `PackageNameError`, using `thiserror`. This error type helps write useful unit tests, `miette` is neat for displaying but not so much for "low-level" errors if that makes sense.
- [Changes `Deref` of `PackageName` to target `str`](https://github.com/helsing-ai/buffrs/pull/132/commits/8bfe6e7fc075e6c67eba6859231103ae9c8e0d4d). In general, similar as with the `TryFrom` implementation, always prefer `&str` over `&String`. The reason is that `&str` is a generic string (can be backed by any container). Here, I just swapped out the `Target` of the `Deref` and made it work.
- [Moves resolving code into `src/resolver.rs`](https://github.com/helsing-ai/buffrs/pull/132/commits/9c4c4c7f482e29cb4b2536e05126194fd1091ccc): in having *one module do one thing*, I felt that resolving was different enough to warrant putting it into it's own module.
- [Derive `Default` for `PackageType`](https://github.com/helsing-ai/buffrs/pull/132/commits/88b3c9e17c3b7433781380ec96ff8593affd4d0b). This type already had a manually-implemented `Default`, but I always prefer `derive` implementations over hand-rolled ones unless absolutely necessary, for maintainability and brevity reasons.
- [Don't use `unwrap_or_default`](https://github.com/helsing-ai/buffrs/pull/132/commits/e54e764410275b1197ba514732dc8c6c30f3935c): Rust has amazing error handling, but there are cases where things actually *should* panic. This is one of those, so I have switched to using `unreachable` rather than silently converting to an empty string. If this ever breaks, we certainly want to know about it!
- [Removes `From<&Package>` implementation](https://github.com/helsing-ai/buffrs/pull/132/commits/39f586bed491ef014a5b11bebd4cab0428cc73d3): this implementation was not used and does not add any value, because the data is available in a public field.
- [Add unit tests for `PackageType`](https://github.com/helsing-ai/buffrs/pull/132/commits/1a56a1f2f17c6cd73b935e65a45becf0dbbd8c48): this cleans up `PackageType` a bit, adding unit tests, making the code a bit clearer.
- [Remove implicit assumptions from `PackageStore`](https://github.com/helsing-ai/buffrs/pull/132/commits/af86bcab5131b8fa93f14640901e75cb5dea4dbf): The `PackageStore` currently heavily relies on the current working directory. I do not like using mutable global state, and also for testability purposes this is not great. So instead I added a field that contains the root path that all relative paths are resolved using.
- [Don't read `Manifest` in `PackageStore::release`](https://github.com/helsing-ai/buffrs/pull/132/commits/2dd5b8b83f74f917783ba0f2d839f2ebf078e723): in the interest of decoupling things, we don't really want the `PackageStore` to have to interact with manifests (such as reading them). Rather, we want to decouple them by having it such that the `Manifest` is a method argument that we simply pass through.
- [Decouples `PackageStore` from compression](https://github.com/helsing-ai/buffrs/pull/132/commits/892256735680da4796e1477cb204880548ac2782): similar as with the `Manifest`, our `PackageStore` should only ever deal with package storage things and not with compression. So I have moved the compression-related code out into `Package` instead, this decouples things.

Before:

test-coverage-before

After:

test-coverage-after

I did not quite get the test coverage up to 80%, but we did hit about 77% (with some cheating because I have moved out the resolver code). 

This PR turned out rather wordy, but I hope it all makes sense! I think in a next iteration I could tackle the testing situation a bit more.

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -72,9 +72,10 @@ pub async fn init(kind: PackageType, name: Option<PackageName>) -> miette::Resul
 
     manifest.write().await?;
 
-    PackageStore::create()
+    PackageStore::create(std::env::current_dir().unwrap_or_else(|_| ".".into()))
         .await
-        .wrap_err(miette!("failed to create buffrs project directories"))
+        .wrap_err(miette!("failed to create buffrs project directories"))?;
+    Ok(())
 }
 
 /// Adds a dependency to this project
@@ -135,14 +136,18 @@ pub async fn remove(package: PackageName) -> miette::Result<()> {
     //     tracing::warn!("failed to uninstall package {}", dependency.package);
     // }
 
-    PackageStore::uninstall(&dependency.package).await.ok(); // temporary due to broken test
+    PackageStore::current()
+        .into_diagnostic()?
+        .uninstall(&dependency.package)
+        .await
+        .ok(); // temporary due to broken test
 
     manifest.write().await
 }
 
 /// Packages the api and writes it to the filesystem
 pub async fn package(directory: impl AsRef<Path>, dry_run: bool) -> miette::Result<()> {
-    let package = PackageStore::release().await?;
+    let package = PackageStore::current().into_diagnostic()?.release().await?;
 
     let path = directory.as_ref().join(format!(
         "{}-{}.tgz",
@@ -191,7 +196,7 @@ pub async fn publish(
 
     let artifactory = Artifactory::new(registry, &credentials)?;
 
-    let package = PackageStore::release().await?;
+    let package = PackageStore::current().into_diagnostic()?.release().await?;
 
     if dry_run {
         tracing::warn!(":: aborting upload due to dry run");
@@ -226,12 +231,11 @@ pub async fn install(credentials: Credentials) -> miette::Result<()> {
             "unexpected error: missing dependency in dependency graph"
         ))?;
 
-        PackageStore::unpack(&resolved.package)
-            .await
-            .wrap_err(miette!(
-                "failed to unpack package {}",
-                &resolved.package.name()
-            ))?;
+        let store = PackageStore::current().into_diagnostic()?;
+        store.unpack(&resolved.package).await.wrap_err(miette!(
+            "failed to unpack package {}",
+            &resolved.package.name()
+        ))?;
 
         tracing::info!(
             "{} installed {}@{}",
@@ -278,7 +282,7 @@ pub async fn install(credentials: Credentials) -> miette::Result<()> {
 
 /// Uninstalls dependencies
 pub async fn uninstall() -> miette::Result<()> {
-    PackageStore::clear().await
+    PackageStore::current().into_diagnostic()?.clear().await
 }
 
 /// Generate bindings for a given language
--- a/src/command.rs
+++ b/src/command.rs
@@ -147,7 +147,11 @@ pub async fn remove(package: PackageName) -> miette::Result<()> {
 
 /// Packages the api and writes it to the filesystem
 pub async fn package(directory: impl AsRef<Path>, dry_run: bool) -> miette::Result<()> {
-    let package = PackageStore::current().into_diagnostic()?.release().await?;
+    let manifest = Manifest::read().await?;
+    let package = PackageStore::current()
+        .into_diagnostic()?
+        .release(manifest)
+        .await?;
 
     let path = directory.as_ref().join(format!(
         "{}-{}.tgz",
@@ -196,7 +200,11 @@ pub async fn publish(
 
     let artifactory = Artifactory::new(registry, &credentials)?;
 
-    let package = PackageStore::current().into_diagnostic()?.release().await?;
+    let manifest = Manifest::read().await?;
+    let package = PackageStore::current()
+        .into_diagnostic()?
+        .release(manifest)
+        .await?;
 
     if dry_run {
         tracing::warn!(":: aborting upload due to dry run");
--- a/src/command.rs
+++ b/src/command.rs
@@ -54,6 +54,7 @@ pub async fn init(kind: PackageType, name: Option<PackageName>) -> miette::Resul
             .to_str()
             .ok_or_else(|| miette!("current directory path is not valid utf-8"))?
             .parse()
+            .into_diagnostic()
     }
 
     let name = name.map(Result::Ok).unwrap_or_else(curr_dir_name)?;
@@ -97,6 +98,7 @@ pub async fn add(registry: RegistryUri, dependency: &str) -> miette::Result<()> 
 
     let package = package
         .parse::<PackageName>()
+        .into_diagnostic()
         .wrap_err(miette!("invalid package name: {package}"))?;
 
     let version = version
--- a/src/command.rs
+++ b/src/command.rs
@@ -55,7 +55,6 @@ pub async fn init(kind: PackageType, name: Option<PackageName>) -> miette::Resul
             .to_str()
             .ok_or_else(|| miette!("current directory path is not valid utf-8"))?
             .parse()
-            .into_diagnostic()
     }
 
     let name = name.map(Result::Ok).unwrap_or_else(curr_dir_name)?;
@@ -100,7 +99,6 @@ pub async fn add(registry: RegistryUri, dependency: &str) -> miette::Result<()> 
 
     let package = package
         .parse::<PackageName>()
-        .into_diagnostic()
         .wrap_err(miette!("invalid package name: {package}"))?;
 
     let version = version
@@ -136,8 +134,7 @@ pub async fn remove(package: PackageName) -> miette::Result<()> {
     //     tracing::warn!("failed to uninstall package {}", dependency.package);
     // }
 
-    PackageStore::current()
-        .into_diagnostic()?
+    PackageStore::current()?
         .uninstall(&dependency.package)
         .await
         .ok(); // temporary due to broken test
@@ -148,10 +145,7 @@ pub async fn remove(package: PackageName) -> miette::Result<()> {
 /// Packages the api and writes it to the filesystem
 pub async fn package(directory: impl AsRef<Path>, dry_run: bool) -> miette::Result<()> {
     let manifest = Manifest::read().await?;
-    let package = PackageStore::current()
-        .into_diagnostic()?
-        .release(manifest)
-        .await?;
+    let package = PackageStore::current()?.release(manifest).await?;
 
     let path = directory.as_ref().join(format!(
         "{}-{}.tgz",
@@ -201,10 +195,7 @@ pub async fn publish(
     let artifactory = Artifactory::new(registry, &credentials)?;
 
     let manifest = Manifest::read().await?;
-    let package = PackageStore::current()
-        .into_diagnostic()?
-        .release(manifest)
-        .await?;
+    let package = PackageStore::current()?.release(manifest).await?;
 
     if dry_run {
         tracing::warn!(":: aborting upload due to dry run");
@@ -239,11 +230,13 @@ pub async fn install(credentials: Credentials) -> miette::Result<()> {
             "unexpected error: missing dependency in dependency graph"
         ))?;
 
-        let store = PackageStore::current().into_diagnostic()?;
-        store.unpack(&resolved.package).await.wrap_err(miette!(
-            "failed to unpack package {}",
-            &resolved.package.name()
-        ))?;
+        PackageStore::current()?
+            .unpack(&resolved.package)
+            .await
+            .wrap_err(miette!(
+                "failed to unpack package {}",
+                &resolved.package.name()
+            ))?;
 
         tracing::info!(
             "{} installed {}@{}",
@@ -290,7 +283,7 @@ pub async fn install(credentials: Credentials) -> miette::Result<()> {
 
 /// Uninstalls dependencies
 pub async fn uninstall() -> miette::Result<()> {
-    PackageStore::current().into_diagnostic()?.clear().await
+    PackageStore::current()?.clear().await
 }
 
 /// Generate bindings for a given language
--- a/src/command.rs
+++ b/src/command.rs
@@ -16,8 +16,9 @@ use crate::{
     credentials::Credentials,
     lock::{LockedPackage, Lockfile},
     manifest::{Dependency, Manifest, PackageManifest, MANIFEST_FILE},
-    package::{DependencyGraph, PackageName, PackageStore, PackageType},
+    package::{PackageName, PackageStore, PackageType},
     registry::{Artifactory, RegistryUri},
+    resolver::DependencyGraph,
 };
 
 #[cfg(feature = "build")]
--- a/src/generator.rs
+++ b/src/generator.rs
@@ -63,7 +63,7 @@ impl Generator {
 
         std::env::set_var("PROTOC", protoc.clone());
 
-        let store = PackageStore::current().into_diagnostic()?;
+        let store = PackageStore::current()?;
         let protos = store.collect(&store.proto_path()).await;
         let includes = &[store.proto_path()];
 
@@ -123,7 +123,7 @@ impl Generator {
     /// Execute code generation with pre-configured parameters
     pub async fn generate(&self) -> miette::Result<()> {
         let manifest = Manifest::read().await?;
-        let store = PackageStore::current().into_diagnostic()?;
+        let store = PackageStore::current()?;
 
         info!(":: initializing code generator");
 
--- a/src/generator.rs
+++ b/src/generator.rs
@@ -12,10 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::{
-    fmt,
-    path::{Path, PathBuf},
-};
+use std::{fmt, path::PathBuf};
 
 use miette::{ensure, miette, Context, IntoDiagnostic};
 use protoc_bin_vendored::protoc_bin_path;
@@ -66,9 +63,9 @@ impl Generator {
 
         std::env::set_var("PROTOC", protoc.clone());
 
-        let store = Path::new(PackageStore::PROTO_PATH);
-        let protos = PackageStore::collect(store).await;
-        let includes = &[store];
+        let store = PackageStore::current().into_diagnostic()?;
+        let protos = store.collect(&store.proto_path()).await;
+        let includes = &[store.proto_path()];
 
         match self {
             Generator::Tonic => {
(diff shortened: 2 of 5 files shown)


src/package.rs
```diff
@@ -13,78 +13,96 @@
...
+    manifest::{self, Manifest, MANIFEST_FILE},
+    registry::RegistryUri,
     ManagedFile,
 };
 
 /// IO abstraction layer over local `buffrs` package store
-pub struct PackageStore;
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PackageStore {
+    root: PathBuf,
```
mara-schulke, line 43: We had a similar PR by @uschi2000

See the comments of #24 

I have to say, i still dislike the oop complexity associated with this, i get that the imperative approach is not maintainable longterm – can we find a functional architecture?

mara-schulke:
Overall i like the split of "resolver" vs "package" – i would strive for as functional / pure designs as possible (i know that is not the case right now due to pragmatism) – but if we refactor stuff, i would want to move away from oop / imperative code as much as sensibly possible

mara-schulke (changes requested):
Overall i like the split of "resolver" vs "package" – i would strive for as functional / pure designs as possible (i know that is not the case right now due to pragmatism) – but if we refactor stuff, i would want to move away from oop / imperative code as much as sensibly possible

src/package.rs
xfbs: How do you define *OOP complexity*?

Can you give an example of how a functional architecture would look like here that would model this better?

src/package.rs
```diff
@@ -13,78 +13,96 @@
...
-        let create = |dir: &'static str| async move {
-            fs::create_dir_all(dir)
+    pub async fn create(path: PathBuf) -> miette::Result<Self> {
+        let store = PackageStore::new(path);
+        let create = |dir: PathBuf| async move {
+            fs::create_dir_all(&dir)
                 .await
                 .into_diagnostic()
-                .wrap_err(miette!("failed to create {dir} directory"))
+                .wrap_err(miette!("failed to create {dir:?} directory"))
```
mara-schulke, line 83:
```suggestion
                .wrap_err(miette!("failed to create {} directory", dir.display()))
```

src/package.rs
mara-schulke: `struct` / _instance_ orientation rather than dataflow orientation

the former results in (potentially huge) interlinked objects, flying around with limited composability where as the latter typically separated data transformation (logic) from io and results in better composability, testability and decomp at the same time

Imo (these functions) where all, mostly unrelated from each other, as they are IO heavy and do very little data transformation. A functional design could be to separate the IO part completely

So in this case, as most of the functions do only IO, i dont see the need for them to "belong" to a struct, rather a module, and remove the functions that nest their IO behavior.

---

As i said on rob's pr:

We also dont really have a need currently for making the path configurable, if thats the case the change is valid, but until then it is unnecessary complexity (imo).

xfbs:
Discussed this today, the resolution here is that I will add a context to the `main.rs` which contains an instance (perhaps an `Arc<PackageStore>`) to avoid instances of this being created all over the place. Most of the code should have access to this context struct and it can be used to add other shared, immutable state.

Besides that, perhaps in a future PR, we should try to make sure that we don't mix pure and non-pure code all over the place. To some extend, this PR already addresses that, but there is still opportunities to improve that in future PRs.

src/command.rs
```diff
@@ -70,9 +72,10 @@ pub async fn init(kind: PackageType, name: Option<PackageName>) -> miette::Resul
 
     manifest.write().await?;
 
-    PackageStore::create()
+    PackageStore::create(std::env::current_dir().unwrap_or_else(|_| ".".into()))
```
mara-schulke, line 75: Can we change the signature so people dont have to do this?

like `PackageStore::current().try_create()` or something like this?

src/command.rs
```diff
@@ -132,14 +136,22 @@ pub async fn remove(package: PackageName) -> miette::Result<()> {
     //     tracing::warn!("failed to uninstall package {}", dependency.package);
     // }
 
-    PackageStore::uninstall(&dependency.package).await.ok(); // temporary due to broken test
+    PackageStore::current()
+        .into_diagnostic()?
```
mara-schulke, line 140: Why is this needed? `PackageStore::xyz` should all return a `miette::Result` which you can just use with `?` directly

src/command.rs
```diff
@@ -188,7 +200,11 @@ pub async fn publish(
 
     let artifactory = Artifactory::new(registry, &credentials)?;
 
-    let package = PackageStore::release().await?;
+    let manifest = Manifest::read().await?;
+    let package = PackageStore::current()
+        .into_diagnostic()?
```
mara-schulke, line 205: Same here

src/command.rs
```diff
@@ -188,7 +200,11 @@ pub async fn publish(
...
+        .release(manifest)
```
mara-schulke, line 206: Can we split the releasing out? As we figured out yesterday its not really related to the package store

mara-schulke: Maybe we can put it into `Package` for now` Like `Package::release` or alike

src/command.rs
xfbs: This code is not from me, it is from @asmello, he said this was a workaround for a broken unit test.

xfbs: Oh wait, I misread. 

No, PackageStore returns a fine-grained error, which in this case is an `std::io::Error`. You cannot use the `?` operator with `miette::Result`. You always need to convert any errors into miette errors work `into_diagnostic()`.

src/command.rs
xfbs: Sure, but that is something that should go into a separate PR. I don't want this one to be too big.

src/generator.rs
```diff
@@ -126,33 +123,32 @@ impl Generator {
...
             .wrap_err(miette!("failed to generate bindings"))?;
 
-        if manifest.package.kind.compilable() {
-            let location = Path::new(PackageStore::PROTO_PATH);
+        if manifest.package.kind.is_compilable() {
             info!(
                 ":: compiled {} [{}]",
                 manifest.package.name,
-                location.display()
+                store.proto_path().display()
```
mara-schulke, line 143: Can we rename the methods to something snappier? `store.dir()` and `store.vendor_dir()`? 

I still think that the "proto_path" is not related to the store, so:

proto/* -> the current package
proto/vendor/* -> the "package store"

So if we would split out the logic for accessing the first path that would be great. I think the weirdness that the vendor dir lives inside of the proto dir creates these blurry borders

src/command.rs
xfbs: I thought we wanted to aim to be functional?

src/package.rs
```diff
@@ -13,78 +13,96 @@
...
+        Ok(Self::new(current_dir()?))
+    }
+
+    /// Open given directory.
+    pub fn open(path: &Path) -> Self {
+        Self::new(path.into())
+    }
+
+    /// Path to the `proto` directory.
+    pub fn proto_path(&self) -> PathBuf {
```
mara-schulke, line 67: Imo this should not return the pathbuf? can we return a borrowed path to prevent an alloc?

src/package.rs
```diff
@@ -173,16 +184,84 @@ impl PackageStore {
...
+            .collect();
+
+        paths.sort(); // to ensure determinism
+
+        paths
+    }
+}
+
+#[test]
+fn can_get_proto_path() {
```
mara-schulke, line 234: can you group all tests that you wrote in `#[cfg(test)] mod test` as usual?

src/package.rs
```diff
@@ -262,88 +317,11 @@ impl PackageStore {
...
-        LockedPackage::lock(self, registry, repository, dependants)
-    }
-}
-
-impl TryFrom<Bytes> for Package {
-    type Error = miette::Report;
-
-    fn try_from(tgz: Bytes) -> Result<Self, Self::Error> {
+    /// Load a package from a precompressed archive.
+    pub fn load(tgz: Bytes) -> miette::Result<Self> {
```
mara-schulke, line 324: Imo lets not do this, also we dont load, but rather parse

src/command.rs
xfbs: Same reason: `current()` only checks some stuff that may return `std::io::Error`. We always try to return the most specific error type if we can.

`miette::Result` does not work like `eyre::Result` or `anyhow::Result`. You cannot just cast any error into a `miette::Error` with `?`, you have to explicitly convert it using `into_diagnostic()`.

We don't want to change the return error of `PackageStore::current()`, because then we would make the errors less useful and we would make testing harder than it needs to be. In my understanding, `miette::Result` is mostly useful for pretty-printing errors.

So we will have to deal with these `into_diagnostic` calls in certain places, as far as I can tell.

src/package.rs
```diff
@@ -385,16 +363,44 @@ impl TryFrom<Bytes> for Package {
...
+        &self,
+        registry: RegistryUri,
+        repository: String,
+        dependants: usize,
+    ) -> miette::Result<LockedPackage> {
+        LockedPackage::lock(self, registry, repository, dependants)
+    }
 }
 
-impl From<&Package> for Bytes {
```
mara-schulke, line 390: Why has this been removed?

src/package.rs
```diff
@@ -429,73 +436,139 @@ impl FromStr for PackageType {
...
+            .find(|(_, c)| !Self::is_allowed(*c));
+        if let Some((index, c)) = illegal {
+            return Err(PackageNameError::InvalidCharacter(c, index));
+        }
 
+        Ok(())
+    }
+
+    /// New package name from string.
+    pub fn new<S: Into<String>>(value: S) -> Result<Self, PackageNameError> {
```
mara-schulke, line 538: Can this go to the top of the impl block?

mara-schulke: And return a miette::Result

src/package.rs
```diff
@@ -429,73 +436,139 @@ impl FromStr for PackageType {
...
-#[derive(Clone, Hash, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Clone, Hash, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Debug)]
 #[serde(try_from = "String", into = "String")]
 pub struct PackageName(String);
 
-impl TryFrom<String> for PackageName {
-    type Error = miette::Report;
+/// Errors that can be generated parsing [`PackageName`][], see [`PackageName::new()`][].
+#[derive(thiserror::Error, Debug, PartialEq)]
+pub enum PackageNameError {
```
mara-schulke, line 482:
```suggestion
enum PackageNameError {
```

mara-schulke: And please, if possible, derive miette diagnostics

src/package.rs
```diff
@@ -429,73 +436,139 @@ impl FromStr for PackageType {
...
-    fn try_from(value: &String) -> Result<Self, Self::Error> {
-        Self::try_from(value.to_owned())
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        Self::new(value)
     }
 }
 
 impl FromStr for PackageName {
-    type Err = miette::Report;
+    type Err = PackageNameError;
```
mara-schulke, line 568:
```suggestion
    type Err = miette::Report;
```

src/generator.rs
xfbs: I stuck with the names that were used before: `PROTO_DIR` and `PROTO_VENDOR_DIR`. I'll create an issue for this.

mara-schulke (changes requested):
So overall, yes we can merge this, but i found a few nits regarding style, and a few issues in terms of unnecessarily moved code, leaking of impl details (see the error related comments, maybe take a look at #95 and #103 for ref) and ergonomic issues (especially the into_diagnostic thing)

src/package.rs
xfbs: It returns the `root` path with a directory appended.

I would not really worry about allocations here, the cost is negligible and changing it would be a premature optimization. But we could add an issue for it for someone to tackle at some point.

Perhaps the package store itself could just contain the path to the vendor dir.

src/package.rs
xfbs: Sure, I'll do that.

src/package.rs
xfbs: It is a red flag to have a `From` for a field unless the field semantically represents that type. A package is a package, not something that is equal to some bytes. Also, most importantly, it was not used. Better to use `package.tgz` explicitly.

src/package.rs
xfbs: I strongly disagree with making this private.

src/package.rs
xfbs: I *heavily* disagree with using `miette::Report` here. `miette` is an error *report* library. It helps you create reports for errors. Meaing: you can use it to convert `low-level` errors into something that is pretty-printable. This is not the place to use it, that is why this PR has removed it. 

The idea is: you use fine-grained errors in your application, and at some point before presenting it to the user, you can turn your error into a report for presentation.

At this point, all we care about is parsing `PackageName`s and returning a useful error type that tells you *exactly why* the parsing failed. Which means an `enum` with fields and an `std::error::Error` implementation.
Downstream, you are then free to turn this into an error *report* using `miette`. But if we have this useful error type, we should use it instead. A `FromStr` should never return a `miette::Report`, that is a big code smell (to me).

src/package.rs
xfbs: I don't quite follow?

src/package.rs
```diff
@@ -13,78 +13,96 @@
...
+        Self { root }
+    }
+
+    /// Open current directory.
+    pub fn current() -> Result<Self, io::Error> {
+        Ok(Self::new(current_dir()?))
+    }
+
+    /// Open given directory.
+    pub fn open(path: &Path) -> Self {
```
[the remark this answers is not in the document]
xfbs: Have created an issue for this.

src/command.rs
```diff
@@ -223,12 +239,11 @@ pub async fn install(credentials: Credentials) -> miette::Result<()> {
...
         ))?;
 
-        PackageStore::unpack(&resolved.package)
-            .await
-            .wrap_err(miette!(
-                "failed to unpack package {}",
-                &resolved.package.name()
-            ))?;
+        let store = PackageStore::current().into_diagnostic()?;
+        store.unpack(&resolved.package).await.wrap_err(miette!(
```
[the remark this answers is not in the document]
xfbs: I don't quite follow?

xfbs:
I've addressed the feedback:
- I've created some issues for some of the issues, because I do not want this PR to get any larger than it already is. Prefer to work in small chunks. 
- Some of the issues I have addressed
- Some of the issues I disagree with, and I have left a comment explaining the rationale.

src/command.rs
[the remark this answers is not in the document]
mara-schulke:
```suggestion
    PackageStore::current().into_diagnostic()?.unpack(&resolved.package).await.wrap_err(miette!(
```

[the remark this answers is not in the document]
mara-schulke: (Formatting is broken)

src/package.rs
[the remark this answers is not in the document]
mara-schulke: I think this should be resolved within this PR if im honest 😅  But okay

src/package.rs
mara-schulke: Its more about semantics than allocations, returning a borrowed path makes the lifetime of the path the same, or shorter, than the store

But also i dont really care

src/package.rs
mara-schulke: So, you moved this from `TryFrom<Bytes>` into `::load` and i think there is no reason, as we dont reuse this, and the semantic meaning of load is weird – we dont do IO here

src/package.rs
mara-schulke: A package is strictly equal to bytes

mara-schulke: So, you can do `Bytes -> Package -> Bytes -> Package -> Bytes` all day

The manifest is only parsed out in the `TryFrom` to prevent the need of ungzipping on every manifest access

src/package.rs
mara-schulke: Please see #95 and #103

mara-schulke: This implementation is public, meaning that you leak this to the outside world

src/package.rs
xfbs: Well, having any kind of complex logic inside of a trait implementation is a strong code smell — so I had to move this into a method. I could make the method non-public and rename it. But I would definitely not leave logic inside the `impl TryFrom<Bytes>`, that is not a good idea.

xfbs: I'll rename it to parse and make it non-pub perhaps.

mara-schulke (approved):

Issue #139, opened 2023-10-20
CI does not run against all crates in the workspace.


xfbs:
Currently, the CI only runs against the `buffrs` CLI. In the future, we will add some supporting crates to `buffrs` that handle things such as validation or act as a self-hostable registry. To ensure that tests are run properly, all workspace members should have be the recipient of CI checks.

PR #140, opened 2023-10-20
Runs CI for all crates in workspace

xfbs:
Currently, only the `buffrs` CLI is the recipient of CI scrutiny. This PR makes it so that all crates are checked.

The way this is done is by adding `--all` or `--workspace` flags to the tools. Examples:

```
cargo test --all
cargo clippy --workspace
cargo fmt --all --check
```

This addresses #139.
One issue that also cropped up and is fixed by this is #141.

### DIFF
--- a/src/credentials.rs
+++ b/src/credentials.rs
@@ -14,7 +14,7 @@
 
 use miette::{miette, Context, Diagnostic, IntoDiagnostic};
 use serde::{Deserialize, Serialize};
-use std::{collections::HashMap, env, path::PathBuf};
+use std::{collections::HashMap, env, io::ErrorKind, path::PathBuf};
 use thiserror::Error;
 use tokio::fs;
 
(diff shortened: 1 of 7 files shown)


src/credentials.rs
```diff
@@ -63,14 +63,22 @@ impl Credentials {
...
-                .wrap_err(ReadError(CREDENTIALS_FILE))?,
-        )
-        .into_diagnostic()
-        .wrap_err(DeserializationError(ManagedFile::Credentials))?;
+        let location = location().into_diagnostic()?;
+
+        // if the file does not exist, we don't need to treat it as an error.
+        let contents = match fs::read_to_string(&location).await {
+            Ok(contents) => contents,
+            Err(error) if error.kind() == ErrorKind::NotFound => Default::default(),
```
mara-schulke, line 71: I disagree with the location of this. I believe you should not change the semantics of `read`.

You should do: `Credentials::read().await.unwrap_or_default()` in the `buffrs::build` script.

For transparency you could use things like `tap` to log out potential errors as cargo warnings. The "Read or default" semantics are really only ever needed in the build script

mara-schulke:
Except for the changed semantics, lgtm

src/credentials.rs
xfbs: We cannot do what you suggested, `Credentials::read().await.unwrap_or_default()` is incorrect as it would mask critical errors such as permissions issues. That is dangerous, because it means you could be in a situation where a credentials file clearly exists, and the tool throws no error when you run it, but the credentials are not used. This leads to debugging hell and I would try to avoid it.

At the same time I think I understand where you are coming from. I believe you don't want to implicitly return a default here.

So instead, I will make a compromise: I will have the `read()` function return a `Result<Option<Self>>`. That way, we can clearly disambiguate between a critical error as `Err(..)` and the file simply not existing as `Ok(None)`. I believe that this is the cleanest way of doing it anyways because it lets the callee decide how to treat a missing file.

mara-schulke: Sounds good! I totally agree with this 😊

mara-schulke (approved):

xfbs:
I've just rebased this due to changes in `main`. Had to fix one issue related to `PackageStore` to make it build.

PR #145, opened 2023-10-20
Implements `buffrs` context.

### LINKED ISSUE #134: Create shared `Context` that is passed around in buffrs CLI.
This shared context should be cheaply clonable (use `Arc` internally). It should also have a `PackageStore` member.

xfbs:
In a discussion it was agreed that the current code base is in a bit of a prototype mode, and some work can be done to set good patterns. We need a `Context` struct that gets passed around which contains all necessary metadata and abstractions that are used for interacting with the `buffrs` project in order to separate concerns and keep code clean.

Currently there are hidden assumptions all over the place, such as implicit state and I/O operations. Any time you call `fs::canonicalize` or `std::env::current_dir`, or you access *anything* with a path that is not an absolute path, you have an **implicit dependency on the current working directory**. The currrent working directory is **mutable, global state**, which can be evil. Similar with reading stuff like the manifest or credentials: we just want to read that data once, at the very beginning, and not multiple times in several places in the code, otherwise we will have an inconsistent view of the world. Instead, we want code that is more compartmentalized and is more *functional*. 

How can we achieve this? We have to think about our data flow. Instead of having a tangled mess of interdependent functions, where everything accesses the outside world, we want to have a clean and clear data flow path. For example, it might look like this:

```mermaid
flowchart LR
    current_dir
    manifest
    credentials
    arguments
    stdout
    side_effects

    subgraph buffrs
        direction LR;
        context
        options
        command
        result
    end

    options-->command
    context-->command
    command-->result
    command-->side_effects
    arguments-->options
    current_dir-->context
    manifest-->context
    credentials-->context
    result-->stdout
```

The idea here is:

- We only ever read stuff *once*, **at the very beginning**, when we build this context. That means this context should contain all information we need to do our job. Specifically, this means the context includes parsed `Credentials`, it includes the root path of our project (and all paths we build, ever, use this root path as base directory). This context represents a snapshot of the information from when the invocation of the `CLI` is started. While the `CLI` runs, data might change on disk and that is fine.
- We only ever write stuff *once*, in the execution of a command. 

## How does it look like in practise?

There is one, and only one instance of `Context`. It is initialized exactly once in the `main.rs`:

```rust
let context = Context::open_current().await?;
```

All subcommands are implemented as methods on this `Context`. The context is wrapped in an `Arc`, so it may be freely passed around, especially to async code that is spawned. It is read-only, and it may never be modified during the duration of the program. `buffrs` is a CLI, so we are fine to read all data once upfront, do our thing, and then exit. We do not need to sync this `Context` with the outside world.

## What does this PR do?

This PR adds a `Context` struct which contains all of that information `buffrs` project. As such, it contains modules used to interact with all parts of a project, including the package store. It also contains metadata such as the credentials that are needed to interacting with the registry.

This PR refactors the code base to use this `Context`.

This resolves #134.

**Note: this PR is the first one of a few, it probably won't get things perfect. I am trying to break tasks down into small, manageable chunks that we can merge rather quickly instead of having large PRs that take a while to review.**

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use crate::{
+    context::Context,
     credentials::Credentials,
     lock::{LockedPackage, Lockfile},
     manifest::{Dependency, Manifest, PackageManifest, MANIFEST_FILE},
@@ -27,7 +28,7 @@ use crate::generator::{Generator, Language};
 use std::path::PathBuf;
 
 use async_recursion::async_recursion;
-use miette::{bail, ensure, miette, Context, IntoDiagnostic};
+use miette::{bail, ensure, miette, Context as _, IntoDiagnostic};
 use semver::{Version, VersionReq};
 use std::{env, path::Path, sync::Arc};
 use tokio::{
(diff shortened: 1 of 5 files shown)


src/command.rs
```diff
@@ -77,255 +77,268 @@ pub async fn init(kind: PackageType, name: Option<PackageName>) -> miette::Resul
...
-        .push(Dependency::new(registry, repository, package, version));
-
-    manifest
-        .write()
-        .await
-        .wrap_err(miette!("failed to write `{MANIFEST_FILE}`"))
-}
+impl Context {
+    /// Adds a dependency to this project
+    pub async fn add(
```
mara-schulke, line 82: Can we inverse this? I think it should take the context, rather than being implemented _on_ the context.

src/command.rs
```diff
@@ -77,255 +77,268 @@ pub async fn init(kind: PackageType, name: Option<PackageName>) -> miette::Resul
...
-        }
+        manifest.write().await
     }
 
-    let artifactory = Artifactory::new(registry, &credentials)?;
-
-    let manifest = Manifest::read().await?;
-    let package = PackageStore::current()?.release(manifest).await?;
+    /// Packages the api and writes it to the filesystem
+    pub async fn package(
```
mara-schulke, line 151: I structutally dislike the change regarding to the commands.rs, it creates a weird asymmetry between different commands

src/context.rs
```diff
@@ -0,0 +1,51 @@
...
+use std::path::PathBuf;
+use std::sync::Arc;
+
+/// Shared context representing `buffrs` project.
+#[derive(Debug)]
+pub struct Context {
+    /// Package store, contains installed dependencies.
+    store: PackageStore,
+    /// Credentials, used to access registries.
+    credentials: Arc<Credentials>,
```
mara-schulke, line 12: If we store an Arc<Context> do we need this to be an arc too?

src/context.rs
```diff
@@ -0,0 +1,51 @@
...
+}
+
+impl Context {
+    /// Package store
+    pub fn store(&self) -> &PackageStore {
```
mara-schulke, line 17: Can we make the fields public rather than using getters?

src/context.rs
```diff
@@ -0,0 +1,51 @@
...
+
+    /// Parsed credentials
+    pub fn credentials(&self) -> &Arc<Credentials> {
+        &self.credentials
+    }
+
+    /// Create new context.
+    ///
+    /// This will create and initialize a new `buffrs` project.
+    pub async fn create<P: Into<PathBuf>>(path: P) -> Result<Arc<Self>> {
```
mara-schulke, line 29:
```suggestion
    pub async fn create<P: Into<PathBuf>>(path: P) -> miette::Result<Arc<Self>> {
```

src/context.rs
```diff
@@ -0,0 +1,51 @@
...
+        Ok(Arc::new(Self {
+            store: PackageStore::create(path.into()).await?,
+            credentials: Credentials::load().await.map(Arc::new)?,
+        }))
+    }
+
+    /// Open a context by path.
+    ///
+    /// This will check if the path is a valid `buffrs` project, and return an error if it is not.
+    pub async fn open<P: Into<PathBuf>>(path: P) -> Result<Arc<Self>> {
```
mara-schulke, line 39: Do we ever do this? I would only implement the api that is currently used

src/main.rs
```diff
@@ -136,9 +135,10 @@ async fn main() -> miette::Result<()> {
         .unwrap();
 
     let cli = Cli::parse();
-    let credentials = Credentials::load().await?;
 
-    match cli.command {
+    // handle init separately, since that is the only command that does not assume an initialized
```
mara-schulke, line 139: I dislike this change, as stated earlier

mara-schulke (changes requested):
So I think the context is implemented sensibly, but I disagree with the (unnecessary) changes made to command.rs

If you revert them and the structural changes implicated by them, this seems reasonable for me and thus good to go

src/context.rs
mara-schulke: Context: If we access this through an arc, we have the guarantee that no one mutates it anyway

src/command.rs
xfbs: The usual way in Rust to structure code like this is by creating contexts and using methods on them. See `reqwest::Client` for example. This allows for dot-chaining (`Context::current().some_method()`), generates useful documentation (*what methods can I call on this context*) and allows for good IDE developer experience (like auto-complete).

Is there a reason we want to do it differently in this project? I can put it into a separate PR if there is some discussion/strategizing that is needed here.

src/command.rs
xfbs: I'm not sure I understand what you mean, can you elaborate?

src/context.rs
xfbs: Yes we can. I find that using getters here is more clean however. This is not a *plain old data* struct that we expect to be stable and not change. We want to be able to change the layout here if we want to without touching a million files.

src/context.rs
xfbs: I added this because I intend to use it for unit tests that are coming in a separate PR.

src/main.rs
xfbs: Please elaborate.

src/command.rs
mara-schulke: I think we loose the notion of having one function per buffrs command, we are very likely to take the context in different places (like eg. - the resolver) aswell and (this is a personal opinion, but maybe relevant:) 
I think the "context" is really just grouped data in memory which you typically want to pass around together, rather than a proper concept which can do things (in your code a context _can_ (semantically) _add_ a dependency) 

Rather than the function that adds a dependency, needing a context to operate in. 

To me these do the same but have two very different semantic properties attached, that's why I marked this. Also it feels odd to have certain commands that don't need a context detached from the rest of the commands. The init command is now in a weird place - and there could be a couple more commands like the init one

src/context.rs
mara-schulke: I think this is not as bad as you make it sound, if we change the name of "store" to something else, we would likely also rename the getter for consistency. 

Bulk renaming is not a bad thing imo - and if we don't want to change the api, we would just not rename anything (is my guess)

src/command.rs
mara-schulke: See my other comment :) The init command is now in a weird place and there are more potential cases for this

src/main.rs
mara-schulke: https://github.com/helsing-ai/buffrs/pull/145/files#r1368292580

src/context.rs
xfbs: It is up to you, but here are some concerns:

- An is not a guarantee for read-only semantics: [`Arc::get_mut()`](https://doc.rust-lang.org/stable/std/sync/struct.Arc.html#method.get_mut).
- This `Context` type is not just a loose bag of data, it has a semantic meaning. It represents one "validated" context of a `buffrs` project. Making the fields public means you can at any point mutate them, or in the case of an `Arc`, override them. Usually, when we build abstractions like this, we don't want to make fields public to prevent the type from being misused. All access should go through explicit member functions and it should only be constructed through explicit constructor methods. 
- As for refactoring: even if the refactoring itself is not so bad, imagine a project that has a healthy number of contributors and about 30 in-progress branches. Now, you need not only refactor your local stuff but you're going to have people yelling at you because all of their work is broken because they need to apply this change in each of those branches. Refactoring is hard to scale, so usually it is easier to have some member methods to serve as the API. There is a saying that summarizes this: *be conservative in the API, but be liberal in the implementation*. By using non-`pub` fields here, we give ourselves the maximum of freedom to change things in the future.

In my personal code, I always differentiate between:

- *plain old data* structs. These typically derive `Serialize` and `Deserialize`. They have all-public fields. The reason that is okay is that these represent external data such as configuration, or API responses. They are guaranteed not to change and the types themselves do not serve as abstractions or encode any information.
- everything else: Implements some kind of an abstraction. All interaction through methods. Never implement `Serialize` and `Deserialize`. 

In fact, this rule is so common that there is a `clippy` lint for it called [`partial_pub_fields`](https://rust-lang.github.io/rust-clippy/master/index.html#/partial_pub_fields), I recommend you read the description for it. I wish it did a better job of explaining the background reasoning of it.

src/command.rs
xfbs: I am happy to invert this if you would like — it is your project after all. 

However, I fundamentally disagree with some of your takes. What I will do is summarize my position and thought process behind this change, that way you have all of the information. If after reading this, you still disagree, then we will do it the way you have proposed.

## Notion of one function per command

> I think we loose the notion of having one function per buffrs command, we are very likely to take the context in different places

I would say, having one function per buffrs command is explicitly an anti-goal. Functions as the primary interface for running `buffrs` are (in my opinion/experience) not the "Rustacean" way of doing things. I would not allow such code to be merged into a project of mine.

This is best explained if you think about this in terms of a programmer who might want to consume `buffrs` as a library, perhaps by building something on top of it. What does this programmer need to do? 

- Open a new `Context`, this can fail if you are not inside a valid `git` repository.
- Find out what you can do with this context (actions as methods)

I will give you a real-life example of this pattern. When you interact with `git` from within Rust:

- You create a new [`Repository`](https://docs.rs/gix/latest/gix/struct.Repository.html), this can fail
- You can find out what actions you can take on this repository because they are exposed as methods

This example in [`gix`](https://docs.rs/gix/latest/gix/struct.Repository.html) is a good example because it resembles closely what we are building here. A low-level abstraction that can be used to interact with a `buffrs` project.

In some languages, [such as Zig (iirc)](https://github.com/ziglang/zig/issues/1170), there is no difference at all between a *method* and a *function*, it is just syntactic sugar. In Rust, there is. It is more of a semantic one, as any method you can also call like `Repository::do_something(&value)` rather than `instance.value()`.

But the difference is: when we define something as a method, semantically it means that this is an action that can be taken on that value. For example, when you *install* a buffrs project, the syntax should be `Context::current().install()`. 

When you use functions, it becomes very difficult to figure out what actions can be taken. This is because in Rustdoc, you cannot filter functions for *all functions that take a value X as parameter*. And the reason you cannot do this is because you do not need to because you are supposed to use `impl` blocks to *associate* functions to values.

So, in my opinion: we specifically don't want to have one *function* per buffrs command, but rather one *method*. It is fine if there are some places where we pass a `Context` by reference. But for the main entrypoint (public API), it should be methods. Functions should not be publicly exposed.

## Grouped data

> I think the "context" is really just grouped data in memory which you typically want to pass around together, rather than a proper concept which can do things. To me these do the same but have two very different semantic properties attached, that's why I marked this.

Everything is fundamentally just *grouped data*, right. But this type is not just a convenience type that groups some arguments so you don't have to repeat them. It carries a significant semantic meaning with it.

If you have an value of type `Context`, it means that you have made sure that at the paths specified within it, there really is a `buffrs` project. You need to do that step to be able to call any action on the `buffrs` project. So, semantically the `Context` states: *at some point in the flow of this program have we made sure that this is in fact a valid `buffrs` project*.

Of course, we are interacting with the filesystem, which is globally shared and mutable state. So we can never be fully certain that you do not delete the folder during the invocation of the command. But our responsibility is to, at least once in the execution, ascertain ourselves that we are in a valid buffrs project so that we can emit a useful error message if this is not the case.

So: in my opinion, this `Context` *is* in fact a proper context with properties that must be upheld. It is an abstraction.

## Init command

> Also it feels odd to have certain commands that don't need a context detached from the rest of the commands. The init command is now in a weird place - and there could be a couple more commands like the init one.

This is not by accident. The code is merely a reflection of the reality.

Using buffrs looks like this:

```mermaid
graph LR
    empty
    empty-->buffrs_init
    buffrs_init-->buffrs_add
    buffrs_init-->buffrs_install
    buffrs_init-->buffrs_remove
    buffrs_init-->buffrs_package
    buffrs_init-->buffrs_publish
    buffrs_init-->buffrs_generate
```

To summarize: there are two distrinct sets of commands here. There are the commands which expect a properly initialized `buffrs` project to be in place (add, install, remove, package, publish, generate, ...) and there is the set of commands that do not (init).

I would perhaps clean this up and make the `init` be an associated function of the `Context` struct and have it return a context. 

## Summary

In my opinion, the changes I made here make sense and overall improve the code base. I also feel that this is not a personal opinion but a style that is widely mirrored in the Rust community. 

If, after this explanation, you agree with these choices we can leave these in. If you disagree, that is okay too. In that case, I will take those changes out of this PR.

src/command.rs
xfbs: I've addressed it in the other comment.

src/context.rs
xfbs: See the `src/resolve.rs`, there are multiple functions which take an `Arc<Credentials>`.

src/main.rs
xfbs: Addressed in the linked comment.

mara-schulke (approved):

kihehs (approved):

xfbs:
I have just rebased this again, and had to change it slightly, because my mental model was somewhat wrong. The actual flow is like this:

```mermaid
graph LR
    empty
    empty-->buffrs_init
    empty-->buffrs_login
    empty-->buffrs_logout
    buffrs_init-->buffrs_add
    buffrs_init-->buffrs_install
    buffrs_init-->buffrs_remove
    buffrs_init-->buffrs_package
    buffrs_init-->buffrs_publish
    buffrs_init-->buffrs_generate
```

Meaning: you should not need to be in a valid buffrs project to be able to use the `login` and `logout` commands. 

By the way, the unit tests in `buffrs` are amazing. I would not have caught this if it were not for them.

PR #147, opened 2023-10-23
Use `Context` in more places

xfbs:
This commit:

- Removes the remaining uses of `PackageStore::current()`
- Passes the `Context` through to the generator module

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -81,7 +81,7 @@ impl Context {
 
     /// Adds a dependency to this project
     pub async fn add(
-        self: Arc<Self>,
+        self: &Arc<Self>,
         registry: RegistryUri,
         dependency: &str,
     ) -> miette::Result<()> {
@@ -125,7 +125,7 @@ impl Context {
     }
 
     /// Removes a dependency from this project
-    pub async fn remove(self: Arc<Self>, package: PackageName) -> miette::Result<()> {
+    pub async fn remove(self: &Arc<Self>, package: PackageName) -> miette::Result<()> {
         let mut manifest = Manifest::read().await?;
 
         let match_idx = manifest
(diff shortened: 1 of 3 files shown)


src/command.rs
```diff
@@ -216,8 +212,61 @@ impl Context {
         artifactory.publish(package, repository).await
     }
 
+    /// Helper method to traverse dependencies and install recursively.
+    #[async_recursion]
+    async fn traverse_and_install(
```
mara-schulke, line 217: Why is this on the ctx?

src/command.rs
```diff
@@ -227,83 +276,34 @@ impl Context {
 
         let mut locked = Vec::new();
 
-        #[async_recursion]
-        async fn traverse_and_install(
```
mara-schulke, line 231: I would leave it in here to make it clear that it is not usable in the outside world

src/command.rs
xfbs: Personal preference. I'm not a big fan of nesting functions inside other functions, I feel like it is not very readable or intuitive.

src/command.rs
xfbs: As it is not `pub`, it is neither visible in docs nor usable outside of this module. As it operates on the current `buffrs` package, I made it a method. Do you prefer the inlined version?

mara-schulke: Yes – they communicate two things, the inlined version communicates that this is only ever relevant to the `install` fn – the private version indicates that this is usable for other commands which it isnt

src/command.rs
mara-schulke: I agree on the readability but i think they communicate an important message to the next engineer working on this.

mara-schulke:
If you could inline it again, it lgtm!

xfbs:
Just inlined it and pushed it!

xfbs:
I'm still not quite happy with this. I find that the method is far too long, and nested too deeply.

I think a good solution at some point that would address both of our issues would be to move the code for this command (maybe even all commands) into separate modules, so that we can have private helper functions. I find that helps a lot with readability.

mara-schulke:
I think the context should be removed again for now though as we have a couple of different subsets of IO that is needed for the commands to work nicely. After spending a day on this i have not gotten a clear image of how this could play together. The current context impl ignores the lockfile completely (which is okay, since it would increase the complexity a lot). See #131

mara-schulke:
Closed by #131 for now

PR #150, opened 2023-10-25
Refactors src/package.rs into multiple modules.

xfbs:
In general, one module should do one thing. The `src/package.rs` module was doing too many things, including:

- compressing and decompressing packages
- handling package stores
- handling package names
- handling package types

Instead, I have split these into small modules that are more obvious in what they do and can be more easily tested. In doing so, I have added some unit tests as well.

In the next step, we can get these modules up to the 80% test coverage. Specifically, `package::compressed` has some low-hanging fruit in terms of refactoring potential (being able to unit test compress-decompress roundtrip is very easily achieved) and `package::store` still has lots of refactoring potential.

The overall goal here is to separate I/O commands from non-I/O commands, so that the latter can be easily unit tested (whereas the former can only be integration tested).

Screenshot 2023-10-25 at 13 47 55


mara-schulke (approved):
Can we move the `package.rs` to `package/mod.rs`?

xfbs:
In general, it is [not recommended not required](https://doc.rust-lang.org/edition-guide/rust-2018/path-changes.html#no-more-modrs) to use `mod.rs` files anymore, instead it is recommended to name them after the name of the module. So `package.rs` is correct whereas `package/mod.rs` is the old style (which is still supported).

The reasoning for that is that it makes it easier to work with the code — you don't end up with a bunch of open `mod.rs` files in your editor.

I don't mind either way, but when in doubt I would stick to the recommendation.

PR #151, opened 2023-10-25
Refactors digest-related code in `src/lock.rs`

xfbs:
This moves the digest-related code out of `src/lock.rs` and adds some tests for it. This increases overall test coverage (line coverage) by about 9%. 

Notes:
- We currently manually implement `Serialize` and `Deserialize`, this is usually not a good idea. It is better (less code) to use something like `serde_with` to implement it.
- There is probably still a couple low hanging fruit but this PR tests the most important stuff.

Before:

Screenshot 2023-10-25 at 14 57 11

After:

Screenshot 2023-10-25 at 14 56 20


mara-schulke (approved):

src/lock/digest.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    fn from_str(input: &str) -> Result<Self, Self::Err> {
+        match serde_typename::from_str(input) {
+            Ok(value) => Ok(value),
+            _other => Err(DigestAlgorithmError::UnsupportedAlgorithm(input.into())),
+        }
+    }
+}
+
+#[test]
```
asmello, line 48: Sorry I'm late to review, but I don't think this is good practice @xfbs. First time I've seen tests outside of a module gated by `#[cfg(test)]` and after [reviewing the docs](https://doc.rust-lang.org/book/ch11-03-test-organization.html#the-tests-module-and-cfgtest) I'm pretty sure there's a good reason for that.

PR #158, opened 2023-10-30
Update package name length constraints

### LINKED ISSUE #156: Package name has to be minimum of three characters. This seems like an arbitrary rule
Is there any reason for why such rule is being enforced?

mara-schulke:
As described #156 the arbitrary constraint of minimum three chars is not suitable for some users. This loosens the constraints to a minimum of one character (ie. non empty) and simultaneously introduces a upper bound of a 128 to prevent abuse.

Closes #156


xfbs (approved):
Looks good to me!

I like that you made the minimum and maximum lengths constants as well.

PR #160, opened 2023-11-03
Removes unused crates

xfbs:
I saw that there are some crates that are not used (thanks to `cargo udeps`), this PR removes them.

Food for thought: maybe we want to add a CI stage for running `cargo udeps`?


mara-schulke (approved):

mara-schulke:
Yes feel free to do so! I would want to keep CI runtime as low as possible though.

PR #163, opened 2023-11-06
Update Getting Started Documentation to reflect current state of implementation

JoeHut:
* Update registry flag usage for login in getting started section
* Update buffrs add usage in "First Steps"

### DIFF
--- a/docs/src/getting-started/first-steps.md
+++ b/docs/src/getting-started/first-steps.md
@@ -46,11 +46,19 @@ your protocol buffers as a package.
 Let us define a dependency of the webserver on a hypothetical library
 called `user` in the `datatypes` repository.
 
-This is done by editing the manifest or by invoking `buffrs add`:
+This is done by invoking `buffrs add`:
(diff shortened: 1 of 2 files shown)


docs/src/getting-started/first-steps.md
```diff
@@ -46,11 +46,19 @@ your protocol buffers as a package.
...
 called `user` in the `datatypes` repository.
 
-This is done by editing the manifest or by invoking `buffrs add`:
+This is done by invoking `buffrs add`:
+
+```bash
+$ buffrs add --registry https://your.registry.com datatypes/user@=0.1.0
+```
+
+The result is a dependency in the `Proto.toml`:
```
mara-schulke, line 56: Can we add a `> Note: Using the exact operator `=` is currently required when specifying dependencies` – maybe phrased better, but to explain why `@=0.1.0` is necessary?

mara-schulke (approved):

PR #164, opened 2023-11-07
Allow code gen when only protos exist

alexespencer:


mara-schulke (approved):

mara-schulke:
@alexespencer can you document this behavior in the book eventually? 
Thank you!

PR #167, opened 2023-11-14
Support releasing of api package from implementing repository

JoeHut:
### ``buffrs install``
Not only the dependencies are fetched and installed, but also the package itself is installed. This allows buffrs to operate on the proto/vendor directory exclusively. In particular, this allows importing .proto files with absolute paths.

### ``buffrs ls``
Output all .proto files under the proto/vendor directory. This allows no compilation of the package and its dependencies through

```sh
protoc --proto_path proto/vendor $(buffrs ls) --python_out out/
```


mara-schulke:
Thank you! I will take a look later today, is that early enough? 😊

mara-schulke:
My only annotation is probably documentation of this behavior in the buffrs guide and the buffrs reference sections of the book

mara-schulke:
Also a couple of things to watch out for:

- How does it behave when run multiple times?
- How well does this change integrate with buffrs lint for example? What are the implications?
- Can we auto populate this, before every buffrs command?
- Implications on existing packages?

JoeHut:
> Thank you! I will take a look later today, is that early enough? 😊

Yeah, sure, no pressure there.


> My only annotation is probably documentation of this behavior in the buffrs guide and the buffrs reference sections of the book

Jep, working on that. Just wanted to get some early feedback whether this makes sense.

JoeHut:
> How does it behave when run multiple times?
The contents in ``proto/vendor/<package>`` are overwritten, so that the current state is synchronized. However, when the ``Proto.toml`` is edited manually (for example, the name of the package is changed), you might end up with orphaned packages. But we have this behavior already, for example if you manually remove a dependency. Should we implement a ``buffrs clean`` command that nukes ``proto/vendor`` and call this before ``buffrs install``?

> How well does this change integrate with buffrs lint for example? What are the implications?
I will have to change this, but to be honest, I am not sure what the expectations there are... From my tests of the currently released version, you would have to have the ``vendor/`` prefix on all imports of libs which are installed as dependencies...

> Can we auto populate this, before every buffrs command?
I will check, but shouldn't then be ``buffrs install`` the command to be auto-populated? Commands like ``buffrs ls`` or ``buffrs lint`` etc. don't really work either as long as ``buffrs install`` is not called, right?

mara-schulke:
> you might end up with orphaned packages. But we have this behavior already, for example if you manually remove a dependency. Should we implement a buffrs clean command that nukes proto/vendor and call this before buffrs install?

There is `buffrs uninstall` for this. But also, i would make sure there are no files / folders in the package store that should not belong there. This is quite easy, as you have all package names of the dependencies + the current package name upon loading the manifest from the `Proto.tom` and can diff it with the file system. 

I would not uninstall everything all the time, just verify that no folders exists that should not exist and remove those which shouldnt be there according to the manifest.

---

> I will have to change this, but to be honest, I am not sure what the expectations there are... From my tests of the currently released version, you would have to have the vendor/ prefix on all imports of libs which are installed as dependencies...

I think this needs research before we can merge it. I think, after populating the vendor folder with the current state of the proto folder – you can just lint `proto/*.proto`. Please make sure it doesnt break 😊 

---

> I will check, but shouldn't then be buffrs install the command to be auto-populated? Commands like buffrs ls or buffrs lint etc. don't really work either as long as buffrs install is not called, right?

Well, no i wouldnt say that. `buffrs install` needs a network connection, the population of the current package is just a copy operation.

src/package/store.rs
```diff
@@ -58,6 +58,9 @@ impl PackageStore {
     /// Open given directory.
     pub async fn open(path: &Path) -> miette::Result<Self> {
         let store = Self::new(path.into());
+        let manifest = Manifest::read().await?;
+
+        store.populate(&manifest).await?;
```
JoeHut, line 63: @mara-schulke I do the sync now here so that it happens every time the store is accessed. Not sure about the manifest however... Is it fine to read it here? Or should it be passed in (seems like touching a lot of signatures)?

JoeHut: Or can we even have the manifest as member of the store?

src/package/store.rs
```diff
@@ -229,6 +234,34 @@ impl PackageStore {
 
         paths
     }
+
+    /// Sync this stores proto files to the vendor directory
+    async fn populate(&self, manifest: &Manifest) -> miette::Result<()> {
```
JoeHut, line 239: @mara-schulke Not sure about all of this here... Is this the right way of error handling with miette?

src/validation/data/package.rs
```diff
@@ -61,13 +61,23 @@ impl Package {
...
             name: descriptor.package().to_string(),
             entities: Default::default(),
         };
 
+        package.parse(descriptor)?;
+
+        Ok(package)
+    }
+
+    pub fn add(&mut self, descriptor: &FileDescriptorProto) {
```
JoeHut, line 74: @xfbs This works but it doesn't feel right... Would be happy to get some opinion here from someone more rusty than me...

xfbs:
Hey @JoeHut, I meant to review this today but I got kind of distracted :smile: so, I'll sent you a review in the morning if that is okay :)

JoeHut:
@xfbs This is not time-critical at all! I am happy if you find the time to have a look :)

src/package/store.rs
mara-schulke: I think that that *could* be sensible - but i would pass it as a function argument for now, as having two instances of the manifest can become problematic in terms of state management (ie. when a dependency is added in first but not the second one)

mara-schulke: My gut feeling is that you want to explicitly populate the store, instead of doing it simultaneously with opening

mara-schulke:
This looks great @JoeHut!! Thank you a lot!

src/validation/data/package.rs
```diff
@@ -61,13 +61,23 @@ impl Package {
...
+        self.parse(descriptor).unwrap();
+    }
+
+    fn parse(&mut self, descriptor: &FileDescriptorProto) -> Result<&Self, PackageError> {
```
mara-schulke, line 78: I think you can inline it in here @JoeHut – so taking a slice of file descriptors, instead of a single one? @xfbs what do you think?

src/package/store.rs
```diff
@@ -145,10 +148,12 @@ impl PackageStore {
         &self,
         manifest: &Manifest,
     ) -> miette::Result<crate::validation::Violations> {
-        let pkg_path = self.proto_path();
-        let source_files = self.collect(&pkg_path, false).await;
+        let root_path = self.proto_vendor_path();
```
mara-schulke, line 151: this looks wrong to me – im 99.9% confident, that we dont want to lint the protocol buffers in `proto/vendor` but the ones in `proto/proto.*`!

mara-schulke: @xfbs what do you think?

src/package/store.rs
mara-schulke: yes, you should ideally add context to the errors after calling `into_diagnostic` using the `WrapErr` trait

src/validation/data/package.rs
mara-schulke: also it is good practice to reduce the number of functions that take an `&mut self` to a minimum since they encourage imperative programing. 

This is essentially a parsing problem which can be solved nicely in a functional manner – so to say "deriving" the contents from the file descriptor types.

So what you want to do (i think):

```haskell
type PackageDescriptors = [FileDescriptor]

parse :: PackageDescriptors -> Package
```

– 

translating this into rust:

```
struct DirectoryDescriptorSet<'a>(&'a [FileDescriptorProto]);

impl<'a> DirectoryDescriptorSet<'a> {
    // implement a mapping function that ensures that the `FileDescriptorProto` structs are located next to eachother in the typesystem
}

struct Package { ... }

impl<'a> TryFrom<DirectoryDescriptorSet<'a>> for Package {
    type Err = miette::Report;

    // implement everything to do type conversion from A (DirectoryDescriptorSet) to B (Package)
    // fn try_from ...
}
```

This way the flow of the program looks like:

1. Collect file descriptors grouped by directory
2. Create DirectoryDescriptorSets from them
3. Try to convert them to one Package using `TryFrom`

cc @xfbs

src/package/store.rs
xfbs: Just to add my 3 cents (inflation): 

I also think that with the state of the code base right now, this could be sensible. However it is not clean because we are violating some rules here:

- Naming: An `open` function does not suggest that it also performs a mutable action.
- Abstraction: We are violating abstraction layers here.

This is not really your fault — we have been violating abstraction layers before. In an ideal world, we would need three types here:

- `LocalPackage`, which owns the files for the current package (`proto`). It has methods to list, read, write files.
- `PackageStore`, which owns the files in the `proto/vendor`. It has methods to list, read, write files.
- `Manifest`, which owns the manifest file.

None of the methods of `PackageStore` should take a `Manifest`, it should not even know that manifests exist. Currently it does, because we are doing quite a bit too much in the `PackageStore`. If you think about abstractions, these low-level types all just handle one particular aspect, but this population action is something that moves things from a `LocalPackage` into a `PackageStore`, so it has to live on a higher abstraction level. So we need some kind of higher-level struct that represents the current Buffrs project:

```rust
pub struct BuffrsProject {
    pub manifest: Manifest,
    pub package: PackageStore,
    pub vendor: VendorStore
}
```

This type would have a `populate()` function that does the right thing. Any and all operations which span multiple parts should be cleanly implemented here, and the underlying types (`PackageStore`, `VendorStore`) should expose methods for interacting with the data (so we don't do `tokio::fs::write(project.package.path(), &[..])` but rather `project.package.write_write("filename", &[...])`).

 I think that right now, it is hard to do this cleanly because the PackageStore violates abstraction layers — it is supposed to only handle interactions with the `proto` folder, but it also has methods that take and use Manifests.
I think if it works the way you are doing it, it should be okay to do it that way. But at some point we should revisit this and see if we can't come up with a nice abstraction here that makes everyone happy.

xfbs:
Just one thing I want to add because we need to be aware of it:

> we want to have a package naming convention like

```
proto
├── root.proto         -> package fancylib
└── sub
    ├── one.proto      -> package fancylib.sub
    └── two.proto      -> package fancylib.sub
```

Previously, there was a one-to-one mapping between *file paths* and *package names* (called an isomorphism, I think). This is nice because it means that if you know that something is in package `mypackage.sub`, there is *one and only one* path it could be in: `mypackage/root.proto`. You do not need to look into the docs or in the repository to figure out what the path is, because it is forced by the package name. This means that only the package names are part of the public API.

With this, we now have an epimorphism (or a surjective homomorphism?), because multiple paths will end up being the same package name. This means that when you have a package name such as `physics.temperature`, you no longer know which file to import. And, depending on the type you want to use, it might be in multiple files. So `physics.temperature.Kelvin` might be stored in `physics/temperature/metric_units.proto`, but `physics.temperature.Fahrenheit` might be stored in `physics/temperature/legacy_units.proto`.

```mermaid
graph LR
    a[proto/a.proto]
    b[proto/b.proto]
    this[proto/sub/this.proto]
    that[proto/sub/that.proto]
    package[package]
    sub[package.sub]
    a-->package
    b-->package
    this-->sub
    that-->sub
```

With this, we need to keep in mind that now, **file names are part of the public API**. This also means that we have to change how we do validation — essentially, before we were storing types basically as `BTreeMap<PackageName, Definitions>, but now we need to store them as `BTreeMap<PackageName, BTreeMap<FileName, Definitions>> and we have to make sure that you don't accidentally define the same type multiple times in multiple files belonging to the same package.

I understand that this scheme is closer to how protocol buffer files are used in practice, I am a bit weary of this change, just because it complicates things:

- It is no longer sufficient to know the package name (`physics.temperature`) and type name (`Kelvin`), you also have to know the file name in order to use it (`physics/temperature/metric_temperatures.proto`). Imagine wanting to use a Rust crate an instead of doing `use semver::Version` you have to know that it is stored in the `semver/version2_impl.rs` file (which should be an implementation detail and not part of the public API!).
- You cannot rename files or move definitions between files. All of those are hard, breaking changes (not to be confused with heartbreaking changes). So basically, when the `semver` package renames the file that they implemented `semver::Version` in to `semver/version_implementation.rs`, they have to bump their major version to `2.0` because this is a breaking change for all users.

I am a little scared of this change! I also don't want to change your minds, just explain the implications of this.

src/package/store.rs
xfbs: Ah, I get what you mean. To be honest — it should not matter, right? As long as they are in sync?

JoeHut: I think we have to make the ``collect`` a bit more accessible...
What we do here is, we validate the files in ``proto/vendor/<package>/*.proto`` with the include path ``proto/vendor/``. Since the validation tries to compile the package, we have to do it there in order to resolve the absolute import paths.

JoeHut:
> This is nice because it means that if you know that something is in package mypackage.sub, there is one and only one path it could be in: mypackage/root.proto. 

You meant to write ``mypackage.sub`` -> ``mypackage/sub.proto``, right?


I get what you mean, you wouldn't be able to do an internal restructuring basically without breaking to the outside.

We could however also just require a public API to be exposed, right?

If you have
```
im-test-lib
├── Proto.toml
└── proto
    ├── api.proto
    ├── mass
    │   └── mass.proto
    ├── temperature.proto
    ├── units.proto
    └── vendor
```

with ``api.proto``

```proto
syntax = "proto3";

package im_test_lib;

import public "im-test-lib/mass/mass.proto";
import public "im-test-lib/temperature.proto";
import public "im-test-lib/units.proto";
```

Consumers would then ``import "im-test-lib/api.proto"`` and the file structure would be a mere internal convenience of the package.

I am sure I am missing something however 😅

JoeHut:
Or as an alternative you can just require ``import public "<new_location>.proto"`` to be used in the original file when moving messages and files. That's in the end what it is there for...

src/package/store.rs
JoeHut: (sorry for resolving the comment, your comment was not auto-loaded, @xfbs !)

I absolutely see your point, but as you said would rather not do the restructuring (at least not in this PR). https://github.com/helsing-ai/buffrs/pull/167/commits/bf79e9eb78eb3921bad93a3a546101e2fa45d3a4
^ I changed here to pass in the manifest and doing the population explicitly. I think it is cleaner in terms of abstraction at the moment, but also more error prone... I am both happy with keeping it and leaving it for now...

xfbs:
Idea:

**We have a strict mapping of package name to file name**.

| Package | File |
| --- | --- |
| mypackage | mypackage/root.proto |
| mypackage.sub | mypackage/sub.proto |
| mypackage.sub.way | mypackage/sub/way.proto |

> but patrick, what about splitting things into multiple files?

**We allow for having project-internal "hidden" files**. These are prefixed with an underscore!

```
im-test-lib
├── Proto.toml
└── proto
    ├── root.proto
    ├── _my_stuff.proto
    ├── mass.proto
    ├── mass
    │   └── _this.proto
    │   └── _that.proto
    ├── temperature.proto
    │   └── _kelvin.proto
    │   └── _fahrenheit.proto
    ├── units.proto
    └── vendor
```

The idea here is that we have a **public API**, whereby there is a simple mapping between package name and file name:

- `im_test_lib` in `root.proto`
- `im_test_lib.mass` in `mass.proto`
- `im_test_lib.temperature` in `temperature.proto`
- `im_test_lib.units` in `units.proto`

But additionally, there are also **implementation details**, which is files that start with an underscore character:

- `mass/_this.proto`, which is re-exported by `mass.proto`
- `mass/_that.proto`, which is re-exported by `mass.proto`
- `temperature/_kelvin.proto`, which is re-exported by `temperature.proto`
- `temperature/_fahrenheit.proto`, which is re-exported by `temperature.proto`

These should not be used externally, they should only be used by the current package to re-export types from into the proper file. The idea here is:

- Anything prefixed with an underscore is like `mod xyz` in Rust: private module (implementation detail, not usable from outside).
- We only lint the public API.
- If our package depends on other packages, we forbid from `import`ing hidden files from other packages (to enforce that you do not use the non-public API).
- Buffrs projects are allowed at any point to reorganize their "hidden" files without it being a breaking API change, so long as the types exported by the public API does not change.

This accomplishes:

1. Freedom to split up files into multiple modules without breaking API
2. Simple mapping of package name to file name, no need to look into the sources to figure out which file exactly defined something
3. Simpler for us to lint, too (we basically only need to lint the public API)

What do you think about this idea, @JoeHut and @mara-schulke?

mara-schulke:
@xfbs we specifically removed this behavior and went to folders mapping to packages! 🙈

xfbs:
@mara-schulke I know! I've moved my comment to a discussion. I'm just thinking out (aloud) what some potentially better approaches might look like (which avoid both the pitfalls of the current approach and the approach we have decided on, I'm not super happy with either of them).

JoeHut:
@xfbs I added the changes we discussed last week.
@mara-schulke Since we enforce now the import through absolute paths, I had to touch the ``registry/proto`` entries to fix the paths. So there is an impact on existing projects with this change...

As a follow up PR I would have a look at how to simplify the parsing as discussed in https://github.com/helsing-ai/buffrs/pull/167#discussion_r1396116573 and make an additional PR for documentation. I think all other comments are addressed now.

mara-schulke (approved):

PR #170, opened 2023-11-17
Make static lifetime of constants explicit

JoeHut:
The most recent version of clippy complains about missing explicit lifetime annotators for constants.

### DIFF
--- a/src/generator.rs
+++ b/src/generator.rs
@@ -53,7 +53,7 @@ pub enum Generator {
 
 impl Generator {
     /// Tonic include file name
-    pub const TONIC_INCLUDE_FILE: &str = "buffrs.rs";
+    pub const TONIC_INCLUDE_FILE: &'static str = "buffrs.rs";
 
     /// Run the generator for a dependency and output files at the provided path
     pub async fn run(&self) -> miette::Result<()> {
--- a/src/package/store.rs
+++ b/src/package/store.rs
@@ -35,9 +35,9 @@ pub struct PackageStore {
 
 impl PackageStore {
     /// Path to the proto directory
-    pub const PROTO_PATH: &str = "proto";
+    pub const PROTO_PATH: &'static str = "proto";
     /// Path to the dependency store
-    pub const PROTO_VENDOR_PATH: &str = "proto/vendor";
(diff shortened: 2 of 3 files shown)


mara-schulke:
As of 1.37 there is a lint in place which encourage omitting this:

https://rust-lang.github.io/rust-clippy/stable/index.html#/redundant_static_lifetimes

Can you give me a reference when this has been changed (and in which version?)

JoeHut:
See the discussion here: https://github.com/rust-lang/rust/issues/115010
tbh, I just saw that https://github.com/helsing-ai/buffrs/pull/167 doesn't pass CI because of this and couldn't reproduce it before a ``rustup update``.

But then again, I don't see "elided_lifetimes_in_associated_constant" in https://rust-lang.github.io/rust-clippy/stable/index.html ?!

xfbs:
I was quite confused by this, let me shine some light here!

> As of 1.37 there is a lint in place which encourage omitting this:
> 
> https://rust-lang.github.io/rust-clippy/stable/index.html#/redundant_static_lifetimes

This is still in place — basically, when you declare a static, you do it like this (without the explicit `'static` lifetime):

```rust
static MY_STATIC: &str = "important stuff here";
```

This has not changed.

> The most recent version of clippy complains about missing explicit lifetime annotators for constants.
> See the discussion here: https://github.com/rust-lang/rust/issues/115010
> tbh, I just saw that https://github.com/helsing-ai/buffrs/pull/167 doesn't pass CI because of this and couldn't reproduce it before a rustup update.

Yes, so crucially, this is not about lifetimes in constants, but lifetimes in **associated constants**. So here, we don't have a simple constant, we have a constant which is associated to a type, like this:

```rust
impl MyType {
    const MY_CONST: &'static str = "abc";
}
```

In this situation, we must give an explicit lifetime at the moment, according to that issue you linked.

> But then again, I don't see "elided_lifetimes_in_associated_constant" in https://rust-lang.github.io/rust-clippy/stable/index.html ?!

Correct, this is not a clippy warning, but rather a compiler warning.

xfbs (approved):
LGTM. 

According to https://github.com/rust-lang/rust/issues/115010, we need to make these changes as they will become a hard error in the future.

PR #172, opened 2023-11-21
Nix support tracking branch

j-baker:


j-baker:
Known bug - at present this does not handle dependencies on crates that themselves use buffrs, when utilised by the Rust integration. Only works with local proto.tomls.

I think one solution would be to revamp the idea of the cache to simply directly reference Nix store paths. And then we specifically interact with Crane's Cargo registry when setting things up for buffrs, looking for proto.toml files in all the downloaded crates.

mara-schulke:
@j-baker we could also remove the feature from the rust integration to download dependencies

I think this causes a lot of trouble for many people unfortunately

mara-schulke:
Also have we verified what the behavior is when crate x depends on buffrs package x and is then published?

Do we bundle the protos inside? I don't think so, my worries here would be that the build of crate x runs a buffrs install through the rust integration. This only works if the consumer has access to the proto repos / registry that the crate uses.. (or am I having a wrong guess here?)

mara-schulke:
Hi @j-baker this branch is a whole 194 commits behind main, is there some way we could backport this or get this merged soon?

mara-schulke:
Closed because main has Nix support since 0.8 (which was based on the work done in this branch)

PR #173, opened 2023-11-25
Implement Registry Database Model

mara-schulke:
## Introduces

This PR introduces the database binding code for the following Models:

- `pub struct User`
- `pub struct UserToken`
- `pub enum PackageType`
- `pub struct Package`
- `pub struct PackageOwner`
- `pub struct PackageInvite`


xfbs (approved):
Very clean, I like it. 

CI is complaining because of missing docs — maybe just allow it in the module? I don't think we care about full documentation at this point, while it implementation is still in flux.

```rust
#![allow(missing_docs)]
```

mara-schulke:
Will do, glad to hear 😊

registry/src/model/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+
+/// Buffrs Package Types
+#[derive(Copy, Clone, Debug, Serialize, Deserialize, PartialEq, Eq, sqlx::Type)]
+#[sqlx(type_name = "package_type", rename_all = "kebab-case")]
+#[serde(rename_all = "kebab-case")]
+pub enum PackageType {
+    /// A library package
+    Library,
+    /// An api package
+    Api,
```
NiklasRosenstein, line 67: Excuse my ignorance if this is obvious for regular Buffrs users: What's the consequence for the registry to be able to differentiate between Api and Library packages?

registry/src/model/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+pub struct UserToken {
+    #[sql(pk)]
+    pub hash: String,
+    #[sql(fk -> User, rename = "user_id")]
+    pub user: i32,
+    #[sql(unique)]
+    pub prefix: String,
+    pub allow_publish: bool,
+    pub allow_update: bool,
+    pub allow_yank: bool,
```
NiklasRosenstein, line 50: Permissions are commonly dealt with using "scopes" jn tokens. It's a bit more permissive for compatibility when introducing or deprecating permission kinds.

```suggestion
    pub scopes: Vec<String>,
```

Corresponding scope names could be `package:yank`, `package:publish`, etc.

registry/src/model/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+/// A token that is identifying a user / used for authentication
+#[derive(Schema, Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[table(schema = "public", name = "user_tokens")]
+pub struct UserToken {
+    #[sql(pk)]
+    pub hash: String,
+    #[sql(fk -> User, rename = "user_id")]
+    pub user: i32,
+    #[sql(unique)]
+    pub prefix: String,
```
NiklasRosenstein, line 47: What's the purpose of this?

registry/src/model/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+    //#[hook(delete = DateTime::now())]
+    pub deleted_at: Option<DateTime<Utc>>,
+}
+
+/// A token that is identifying a user / used for authentication
+#[derive(Schema, Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[table(schema = "public", name = "user_tokens")]
+pub struct UserToken {
+    #[sql(pk)]
+    pub hash: String,
```
NiklasRosenstein, line 43: What kind of hash is this/what is it used for? A JWT signature?

registry/src/model/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+    #[sql(fk -> User, rename = "user_id")]
+    pub user: i32,
```
NiklasRosenstein, line 45: Why break naming consistency between the schema and the code?

```suggestion
    pub user_id: i32,
```

registry/src/model/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+    #[sql(unique)]
+    pub prefix: String,
+    pub allow_publish: bool,
+    pub allow_update: bool,
+    pub allow_yank: bool,
+    #[sql(timestamp = created)]
+    pub created_at: DateTime<Utc>,
+    #[sql(timestamp = updated)]
+    pub updated_at: DateTime<Utc>,
```
NiklasRosenstein, line 54: Do tokens get updated?

registry/src/model/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// An api package
+    Api,
+}
+
+/// A package tracked by the registry
+#[derive(Schema, Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[table(schema = "public", name = "packages")]
+pub struct Package {
+    #[sql(pk)]
+    pub id: i32,
```
NiklasRosenstein, line 75: Since the name is unique, why have the additional ID?

registry/src/model/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+#[table(schema = "public", name = "package_owners")]
+pub struct PackageOwner {
+    #[sql(pk)]
+    pub id: i32,
+    //#[sql(fk -> User, rename = "user_id")]
+    #[sql(rename = "user_id")]
+    pub user: i32,
+    //#[sql(fk -> Package, rename = "package_id")]
+    #[sql(rename = "package_id")]
+    pub package: i32,
```
NiklasRosenstein, line 97: Same as above on naming consistency 

I see a few more of these so I'll stop commenting this from here on.

registry/src/model/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+}
+
+/// A package invite
+#[derive(Schema, Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[table(schema = "public", name = "package_invites")]
+pub struct PackageInvite {
+    #[sql(pk)]
+    pub id: i32,
+    #[sql(unique)]
+    pub token: String,
```
NiklasRosenstein, line 115:
```suggestion
    pub token_hash: String,
```

I assume this is the foreign key on the token, not the encoded token.


Separately, could you elaborate on why the package invite requires a token reference? In conjunction some description of this invite process may be useful/good as a docstring!

NiklasRosenstein (changes requested):

registry/src/model/mod.rs
mara-schulke: It's renamed for the purpose of atmosphere generating queries.

I.e. it generates a `UserToken::user` function that loads the `User` with the respective primary key from the db - so this function is derived from the rust field. If we use user_id instead of user the function name doesn't really make sense!

registry/src/model/mod.rs
mara-schulke: Maybe be a copy and paste error

registry/src/model/mod.rs
mara-schulke: Good point

mara-schulke:
cc @xfbs for the Schema! 

I mostly wrote the rust bindings here. I agree that we should fix a couple of things about it 😊

mara-schulke:
@qsantos do you have resources to take this over?

registry/src/model/mod.rs
mara-schulke: It's package validation. See the buffrs book: https://helsing-ai.github.io/buffrs/guide/package-types.html

registry/src/model/mod.rs
mara-schulke: I disagree with `Vec<String>` but it should be `Vec<Scope>` and `enum Scope { Yank, Publish, ... }`

registry/src/model/mod.rs
mara-schulke: cc @xfbs

registry/src/model/mod.rs
mara-schulke: cc @xfbs

registry/src/model/mod.rs
qsantos: I actually like having an `id` column in all my fields. It's good for consistency. It makes things more convenient for joins. It is also more normalized without overdoing it.

Regarding normalization, this is especially useful if the other unique field changes? Although it might be contributed, take this for instance: name normalization changes, making `some_package` be named `some-package` internally. Having an `id` for internal references allows us to not have to think about what could happen to `name`.

registry/src/model/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+
+/// A package owner
+#[derive(Schema, Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[table(schema = "public", name = "package_owners")]
+pub struct PackageOwner {
+    #[sql(pk)]
+    pub id: i32,
+    //#[sql(fk -> User, rename = "user_id")]
+    #[sql(rename = "user_id")]
+    pub user: i32,
```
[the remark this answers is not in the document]
qsantos: It's less important for many-to-many tables, but it does not hurt. It makes things consistent and, more importantly, it's one fewer thing to think about: just put `id` everywhere.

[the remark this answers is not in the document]
qsantos: I wonder about the `rename` though. Why not just keep the column name to `user_id`. That might help make the code more readable as well.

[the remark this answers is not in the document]
mara-schulke: This has to do with the queries that atmosphere generates

It generates functions like "async fn <col>()" and I found "Package::user() -> User" to be the better function name to load the user model than "Package::user_id() -> User"

May be worth to redesign this in atmosphere!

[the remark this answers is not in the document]
mara-schulke: For now resolved, I need to update atmosphere

registry/src/model/mod.rs
mara-schulke: argon2 hash (params noted) of the randomly generated token

mara-schulke:
Re: id columns @NiklasRosenstein I would favor having them over using things like names etc. Using unique names could make renaming very hard, even if they are supposed to be static.

registry/src/model/mod.rs
qsantos: PHC format?

PR #183, opened 2023-12-06
Clear the proto store before installing dependencies

### LINKED ISSUE #100: Clear vendors dir before installing
To make sure installs start from a clean state and avoid conflicts across upgrades.

mara-schulke:
closes #100

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -244,6 +244,7 @@ pub async fn install() -> miette::Result<()> {
     let store = PackageStore::current().await?;
     let credentials = Credentials::load().await?;
 
+    store.clear().await?;
     store.populate(&manifest).await?;
 
     let dependency_graph =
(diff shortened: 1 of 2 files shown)


src/command.rs
```diff
@@ -244,6 +244,7 @@ pub async fn install() -> miette::Result<()> {
     let store = PackageStore::current().await?;
     let credentials = Credentials::load().await?;
 
+    store.clear().await?;
```
tomkarw, line 247: Shouldn't this fail linting? There's now an unused `Option`.

I think the intent would be clearer if we did
```
let _ = store.clear().await;
```

Better yet, let's not silence errors and print a brief message about it, something like:
```
if let Err(err) = store.clear().await {
    warn!("failed to clear store: ", ?err);
}
```
Unless this is very common and would just be noise.

tomkarw (approved):

src/command.rs
mara-schulke: There is no unused option?

PR #188, opened 2023-12-07
Add docs for absolute import paths and vendor copies

JoeHut:


JoeHut:
@mara-schulke Not sure whether this is a good place to document the behaviour. Especially the part about the absolute paths might make more sense somewhere more prominent?!

docs/src/guide/project-layout.md
```diff
@@ -23,7 +23,79 @@ physic
 This will create the `Proto.toml` file which is the manifest file that buffrs
 uses. The `proto` directory, which is the source directory for all your protocol
 buffer definitions and the `proto/vendor` directory, which contains external
-protocol buffers.
+protocol buffers. Before compilation, a copy of the locally defined definitions
```
mara-schulke, line 26: I would not talk about this, as the vendor folder should not be interesting to the user.

docs/src/guide/project-layout.md
```diff
@@ -23,7 +23,79 @@ physic
...
+is automatically copied to `proto/vendor` as well under a directory named
+according to the project name in `Proto.toml`.
+
+A project containing this ``Proto.toml``:
```
mara-schulke, line 30: As said above; to detailed imo

docs/src/guide/project-layout.md
```diff
@@ -23,7 +23,79 @@ physic
...
+├── Proto.toml
+└── proto
+    ├── root.proto
+    ├── length.proto
+    └── calculations
+        ├── distance.proto
+        └── graph.proto
+```
+
+would end up for compilation after installation as:
```
mara-schulke, line 56: It's actually not after compilation, but every single buffrs command

mara-schulke (commented):
I would rephrase the docs to be less implementation related and more user oriented. Answering the question of "how do I do self referencing imports"

(I would create a separate page in the buffrs guide e.g. "import system")

mara-schulke (approved):

JoeHut:
🥳 Thank you :)

PR #190, opened 2023-12-07
Add a Cargo.lock to the registry crate

qsantos:
This PR:

- adds `registry/Cargo.lock`, which was ignored, to the repository
- remove unused dependencies (including `sqlx`, which will need to be re-added when it is actually used); this will make compilation faster until these dependencies are actually needed
- ~~make the S3 storage feature non-default; this should make compilation faster assuming we generally just want a filesystem storage~~
- ignores the `proto/vendor/` directory which is generated by `build.rs` 

If you think any of these should not be done, I'll just remove the corresponding commit.

### DIFF
--- a/registry/.gitignore
+++ b/registry/.gitignore
@@ -1,0 +1,1 @@
+proto/vendor/


mara-schulke:
We want s3 by default!

mara-schulke (approved):
Looks great, just the default feature is important before merging

PR #194, opened 2023-12-10
Deprecate Impl packages (again)

mara-schulke:
## Breaking Changes

Removes `impl` packages

## Follow Ups

We should introduce manifest versions as @jonhoo and @j-baker suggested as they introduce a way to introduce breaking changes to the manifest without risking bad error messages for the user.


xfbs:
Just for my clarity, here's how I understand this:

Buffrs used to support three kinds of packages:

- `lib`, which contain `message` definitions (including enums, and all the other good stuff)
- `api`, which (only?) contain `service` definitions
- `impl`, which was intended to contain code to enhance the lib definitions (?)

Furtheremore, there is a restriction on how dependencies can work, in that packages can only depend on `lib` packages, and not on `api` packages:

```mermaid
graph BT
    foundation_lib
    my_lib-->foundation_lib
    other_lib-->foundation_lib
    my_service_api-->my_lib
    my_service_api-->other_lib
    other_api-->other_lib
```

This PR removes the third category of buffers packages from both the code and the documentation, for two reasons:

- We should use protobuf annotations to change which traits are implemented for types
- We don't want to ship code as packages, because that would mean having to maintain it in multiple languages, exactly what protobuf is trying to avoid.

Is that an accurate representation of the intention of this PR?

mara-schulke:
> We should use protobuf annotations to change which traits are implemented for types

I don't get this point? We never shipped the impl packages! They were not publishable, which is a useless package type imo. This is why I removed them again.

> We don't want to ship code as packages, because that would mean having to maintain it in multiple languages, exactly what protobuf is trying to avoid.

Also not really – as above. They were just there for potetial future telemetry and to mark a project as "dead end" – eg. not producing a package. 

I remove the type to keep the `[package]` section exclusively for projects that are able / intended to publish a buffrs package

xfbs:
> I don't get this point? We never shipped the impl packages! They were not publishable, which is a useless package type imo. This is why I removed them again.

Please don't understand me wrong, I'm just trying to understand what the original intention for the `impl` packages was, and not trying to argue for or against them :smile: 

> They were just there for potential future telemetry and to mark a project as "dead end" – eg. not producing a package.

Gotcha! I think I had mistakenly assumed that the `impl` is short for implementation and that it somehow might contain code (that could be published), my bad!

> I remove the type to keep the `[package]` section exclusively for projects that are able / intended to publish a buffrs package

Ahh okay — so is this a better summary:

- Anything that uses `buffrs` needs to have a `Buffrs.toml` file.
- Some projects just consume buffrs packages, but are not buffrs packages themselves.
- There are some fields in the `Buffrs.toml` file that are mandatory (such as the `[package]` section)
- The `impl` type existed to tell buffrs: "this thing is not a buffrs package, it just depends on buffrs packages"

So the solution is to remove that option, and instead make the whole `[package]` section optional, so that:

- Buffrs packages need to have a `[package]` section in their `Buffrs.toml`
- Projects which consume buffrs have a `Buffrs.toml` which does not contain a `[package]` section, only a `[dependencies]` section
- Now the presence of a `[package]` section indicates that this project is something which publishes a buffrs package, and the type can be one of `lib` or `api`.

Is that a better summary?

mara-schulke:
s/Buffrs.toml/Proto.toml/g

mara-schulke:
Exactly @xfbs 😊 

I also updated the docs. Re-reading the guide _should_ make this clearer now.

Also @j-baker I think we can use @uschi2000's local store impl for the cache that you need for nix.

xfbs:
Sweet, that makes a lot of sense!

docs/src/guide/package-types.md
```diff
@@ -1,38 +1,87 @@
...
-they cannot be referenced as dependencies in packages. However, they do define
-a package type that can depend on both library and API packages. They are final
-consumers of reusable protocol buffers and contain the physical implementation
-of services, or clients and servers -- so they are the intended target of the
-[`generate`](../commands/buffrs-generate.md) command.
+Buffrs makes distinctions between two different packages:
+
+```toml
+[package]
+type = "lib" | "api"
```
xfbs, line 7: I would consider writing this as:

```toml
[package]
type = "lib" # or "api"
```

Because that way, syntax highlighting is not broken (at least on GitHub) 👵🏼

src/command.rs
```diff
@@ -39,7 +39,7 @@ const INITIAL_VERSION: Version = Version::new(0, 1, 0);
 const BUFFRS_TESTSUITE_VAR: &str = "BUFFRS_TESTSUITE";
 
 /// Initializes the project
-pub async fn init(kind: PackageType, name: Option<PackageName>) -> miette::Result<()> {
+pub async fn init(kind: Option<PackageType>, name: Option<PackageName>) -> miette::Result<()> {
```
xfbs, line 42: Would it make sense to have separate `init_package` and `init_consumer` here, due to the different intent and manifests they write? It can quickly get unwieldy when we have a lot of optional arguments which change the semantics of the function (can close this, just an idea).

src/validation/rules/package_name.rs
```diff
@@ -18,33 +18,27 @@ use super::*;
...
-}
-
 impl Rule for PackageName {
     fn rule_info(&self) -> &'static str {
         "Make sure that the protobuf package name matches the buffer package name."
     }
 
     fn check_package(&mut self, package: &Package) -> Violations {
-        if !is_prefix(&self.name, &package.name) {
+        let transposed = self.name.to_string().replace('-', "_");
```
xfbs, line 37: A few notes here:

- We do not need to do the replacement of `-` to `_` here, because the `is_prefix` does that for us
- I have named the `is_prefix` rather poorly because the name does not reflect what it does
- We are kind of violating some abstractions here, I think ideally `crate::package::PackageName` should have a `.transposed()` method that is the canonical way of converting buffrs package names to protocol buffer package names, and we should not hardcode it here (that is on me because I wrote this code :smile:)

Feel free to close this but I wanted to note it down and maybe get back to it at some point.

src/validation/rules.rs
```diff
@@ -100,6 +103,9 @@ impl Rule for RuleSet {
 }
 
 /// Get default rules for a given `buffrs` package name.
-pub fn package_rules(name: &str) -> RuleSet {
-    vec![Box::new(PackageName::new(name)), Box::new(IdentCasing)]
+pub fn all(manifest: &PackageManifest) -> RuleSet {
```
xfbs, line 106: I love this change, because it means that we could at some point allow you to configure rules in the `Proto.toml` file, similar to how cargo lets you do it now:

```toml
[lints.rustc]
missing_docs = "warn"
```

src/package/name.rs
```diff
@@ -50,6 +50,11 @@ impl PackageName {
         Ok(Self(value))
     }
 
+    /// New package name from an unchecked string.
```
xfbs, line 53: Having this means we can violate "type safety". Are there multiple places where we are using this?

I think I only saw one use of it, where we do: `PackageName::unchecked(".")`. If that is the only use, do you think it might be a good strategy to implement `Default` instead and then call `PackageName::default()` there? 👵🏼

src/manifest.rs
```diff
@@ -75,7 +75,7 @@ pub type DependencyMap = HashMap<PackageName, DependencyManifest>;
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Manifest {
     /// Metadata about the root package
-    pub package: PackageManifest,
+    pub package: Option<PackageManifest>,
```
xfbs, line 78: Would it perhaps make sense to have a separate `PublishableManifest` type which has a non-optional `PackageManifest`? Something like:

```rust
impl Manifest
    fn publishable(&self) -> Option<PublishableManifest>;
}
```

And then all places where we need a manifest and expect it to have a `[package]` section, this is the type it will take. Not sure if this makes sense :)

tomkarw:
Why is there a constraint on `lib` packages that they can't depend on other `lib` packages? Is there a good reason behind it, or is it a leaked implementation detail (and if so, it'd be nice to add it [to the lib definition section](https://helsing-ai.github.io/buffrs/guide/package-types.html#libraries)?

In other ecosystems one has `libraries` and `binaries` and libs can depend on libs, bins can depend on libs and nothing can depend on bins. Would such boundaries make sense for Proto files? `lib`s define types (`Message`s) and depend on other `lib`s and `api`s depend on `lib`s and define `Services`.

Then again, I think this PR implies there is no reason to publish `Service` definitions, as it's not something reusable? So we could simplify to just `lib`s then?

xfbs:
> Why is there a constraint on lib packages that they can't depend on other lib packages?

Is there such a constraint? My understanding is that this should not be the case. 

> this PR implies there is no reason to publish Service definitions, as it's not something reusable?

*Au contraire* : there is a good reason to publish service definitions, that is the `api` package type. They have only one restriction: `lib` packages cannot depend on them. 

> In other ecosystems one has libraries and binaries and libs can depend on libs, bins can depend on libs and nothing can depend on bins. Would such boundaries make sense for Proto files? libs define types (Messages) and depend on other libs and apis depend on libs and define Services.

This is exactly the intention here. I think this is a good design, it nudges one towards creating reusable type definition in a separate `lib` package, and keep `api` packages mean and lean.

tomkarw:
> > Why is there a constraint on lib packages that they can't depend on other lib packages?
> 
> Is there such a constraint? My understanding is that this should not be the case.
> 

Auch, my bad. I misread this part of the docs (`can have no ... != can't have ...`)
```
Library packages are the most basic kind of package, as they only define primitive types and can have no dependencies.
```

> > this PR implies there is no reason to publish Service definitions, as it's not something reusable?
> 
> _Au contraire_ : there is a good reason to publish service definitions, that is the `api` package type. They have only one restriction: `lib` packages cannot depend on them.
>

Again, my bad. I forgot that in gRPC `Service` definitions are shared between the server and the client.

> > In other ecosystems one has libraries and binaries and libs can depend on libs, bins can depend on libs and nothing can depend on bins. Would such boundaries make sense for Proto files? libs define types (Messages) and depend on other libs and apis depend on libs and define Services.
> 
> This is exactly the intention here. I think this is a good design, it nudges one towards creating reusable type definition in a separate `lib` package, and keep `api` packages mean and lean.

Going back to this comment:

> Just for my clarity, here's how I understand this:
> 
> Buffrs used to support three kinds of packages:
> 
>     * `lib`, which contain `message` definitions (including enums, and all the other good stuff)
> 
>     * `api`, which (only?) contain `service` definitions
> 
>     * `impl`, which was intended to contain code to enhance the lib definitions (?)

My current understanding is that `impl` was necessary for codebases that themselves didn't define any messages or services, but pulled in `api` definitions and implemented the logic for it (for server).

For example:
```mermaid
graph BT
    some-server-impl
    some-client-impl
    some-api-->some-server-impl
    some-api-->some-client-impl
    lib-1-->some-api
    lib-2-->some-api
    lib-3-->lib-1
    lib-3-->lib-2
```

The `impl` package type would then be necessary to allow depending on `api` and disallowing pushes (`impl` is inherently not publishable).

Also, I wouldn't disallow users to define messages in `api` packages. It'd be preferable, but granularity of composability should be up to the user.

xfbs:
> My current understanding is that impl was necessary for codebases that themselves didn't define any messages or services, but pulled in api definitions and implemented the logic for it (for server).

Yep! I figured that out earlier too. :smile:

> Also, I wouldn't disallow users to define messages in api packages. It'd be preferable, but granularity of composability should be up to the user.

Fair point! I would agree with you.

docs/src/guide/package-types.md
```diff
@@ -1,38 +1,87 @@
...
+    ..
+  }
+}
+```
+
+### `api` – APIs
+
+APIs are the next logical building block for real world systems – they define
+services and RPCs that your server can implement. You can use the
+aforementioned libraries to fuel your development / api definition experience.
```
mara-schulke, line 54:
```suggestion
aforementioned libraries to fuel your development / API definition experience.
```

src/manifest.rs
mara-schulke: Well, you can only create a `Package` if the `PackageManifest` is found in the manifest

mara-schulke:
> Also, I wouldn't disallow users to define messages in api packages. It'd be preferable, but granularity of composability should be up to the user.
This is not planned fyi

tomkarw (approved):
It'd be nice to extend the documentation with content that would answer questions me and @xfbs asked here, but it shouldn't be a blocker for merging this.

mara-schulke:
I ageee @tomkarw, I'm just pretty busy atm / need to get this out asap

src/validation/rules/package_name.rs
mara-schulke: We need it for logging

mara-schulke: I agree, I just don't have the time right now

PR #199, opened 2023-12-14
Add full test of the typical use-case

qsantos:
This MR adds a new command `test-registry`, which can be used for local testing by emulating a (very) dummy registry. This is then used to add a test for the `publish` command, and another one for an updated version of the [tutorial](https://blog.helsing.ai/buffrs-a-package-manager-for-protocol-buffers-2-2-9f001faef6dc).

This brings the total line coverage from 51.30% (1323/2579) to 70.15% (1838/2620).

Note that this MR is best reviewed commit-by-commit.


src/command.rs
```diff
@@ -434,6 +434,60 @@ pub async fn logout(registry: RegistryUri) -> miette::Result<()> {
...
+};
+use bytes::Bytes;
+use std::collections::HashMap;
+use std::sync::{Arc, RwLock};
+
+type State = Arc<RwLock<HashMap<String, Bytes>>>;
+
+/// Run a minimal registry for local testing
+pub async fn test_registry(listen: SocketAddr) -> miette::Result<()> {
+    let state = Arc::new(RwLock::new(HashMap::<String, Bytes>::new()));
```
tomkarw, line 452: All of this should be wrapped in:
```
#[cfg(test)]
mod test_registry {
   ...
}
```

As it shouldn't be compiled when releasing actual binary.

src/main.rs
```diff
@@ -127,6 +128,13 @@ enum Command {
         #[clap(long)]
         registry: RegistryUri,
     },
+
+    /// Run a minimal registry for local testing
+    TestRegistry {
+        /// Address to listen to for incoming connections.
+        #[clap(long, short, default_value = "127.0.0.1:4367")]
+        listen: SocketAddr,
+    },
```
tomkarw, line 137: Same as above.

Cargo.toml
```diff
@@ -48,14 +48,15 @@ serde = { version = "1", features = ["derive"] }
...
-tokio = { version = "^1.26", features = ["fs", "rt", "macros", "process", "io-std", "tracing"] }
+tokio = { version = "^1.26", features = ["fs", "rt", "macros", "process", "io-std", "tracing", "rt-multi-thread"] }
 toml = "0.8.0"
 tonic-build = { version = "0.10.0", optional = true }
 tracing = "0.1"
 tracing-subscriber = "0.3"
 url = { version = "2.4", features = ["serde"] }
 walkdir = "2"
 sha2 = "0.10.8"
+axum = { version = "0.7.2", default-features = false, features = ["tokio", "http1"] }
```
tomkarw, line 59:
```suggestion
[dev-dependencies]
axum = { version = "0.7.2", default-features = false, features = ["tokio", "http1"] }
```

Cargo.toml
```diff
@@ -48,14 +48,15 @@ serde = { version = "1", features = ["derive"] }
 serde_typename = "0.1"
 tar = "0.4"
 thiserror = "1.0.49"
-tokio = { version = "^1.26", features = ["fs", "rt", "macros", "process", "io-std", "tracing"] }
+tokio = { version = "^1.26", features = ["fs", "rt", "macros", "process", "io-std", "tracing", "rt-multi-thread"] }
```
tomkarw, line 51:
```suggestion
[dev-dependencies]
tokio = { version = "^1.26", features = ["fs", "rt", "macros", "process", "io-std", "tracing", "rt-multi-thread"] }
```

src/main.rs
```diff
@@ -228,5 +236,8 @@ async fn main() -> miette::Result<()> {
         Command::Generate { language, out_dir } => command::generate(language, out_dir)
             .await
             .wrap_err(miette!("failed to generate {language} language bindings")),
+        Command::TestRegistry { listen } => command::test_registry(listen)
```
tomkarw, line 239: You get the point :)

tomkarw (changes requested):
Love the effort, automated e2e test is a must for a project like that.

Only feedback is to put the all of  test functionality in `#[cfg(test)]`, as to compile it only conditionally.

src/command.rs
qsantos: I am not quite sure. I intended `test-registry` for users who are following a tutorial or want to test the publish/add commands without actually publishing packages to a a real registry.

If @mara-schulke and you think it's better not to add such a command, i can remove it entirely, and move the `test_registry` function to `test/`.

mara-schulke: Yes, there should not be such a command for the user. 

I am questioning the usefulness of this registry tho! You can't properly use buffrs anyway until you have a artifactory instance and for e2e testing we use something different.

mara-schulke:
So my take on this is, that while this is a useful tool for you, I don't know whether this serves any use case beyond your experiments with buffrs – this it may be best suited in a private binary / a small repo of your private account? I don't think that this registry will be sensibly usable for e2e testing / local experimentation. We do have a registry implementation ongoing (cc @xfbs) that we are planing on using for the e2e tests, so they overlap in scope.

tests/cmd/init/api/mod.rs
```diff
@@ -12,8 +12,7 @@ fn fixture() {
         .assert()
         .success()
         .stdout(include_str!("stdout.log"))
-        .stderr(include_str!("stderr.log"))
-        .code(0);
```
mara-schulke, line 16: why remove the code expectation?

tests/cmd/init/lib/mod.rs
```diff
@@ -12,8 +12,7 @@ fn fixture() {
         .assert()
         .success()
         .stdout(include_str!("stdout.log"))
-        .stderr(include_str!("stderr.log"))
-        .code(0);
```
mara-schulke, line 16: cc, all

tests/cmd/publish/mod.rs
```diff
@@ -0,0 +1,21 @@
+use crate::{with_test_registry, VirtualFileSystem};
+
+#[test]
+fn fixture() {
+    with_test_registry(|url| {
+        let vfs = VirtualFileSystem::copy(crate::parent_directory!().join("in"));
+        crate::cli!()
```
mara-schulke, line 7:
```suggestion

        crate::cli!()
```

mara-schulke:
Okay, so after re-reading everything, I think you are indeed adding a lot of good tests here. Feel free to keep this here, even tho I think it should be replaced by the proper registry longterm.

mara-schulke:
Are _all_ commands tested? That would be great because then we could finally close the issue that tracks that

src/command.rs
tomkarw: Citing Mara:
> Yes, there should not be such a command for the user.

plus

> Okay, so after re-reading everything, I think you are indeed adding a lot of good tests here. Feel free to keep this here, even tho I think it should be replaced by the proper registry longterm.

I think the consensus is to keep everything, but put it behind `#[cfg(test)]`.

qsantos: > You can't properly use buffrs anyway until you have a artifactory instance

That's why I made this command. I wanted to be able to follow along the tutorial without having to upload packages to the artifactory.

I'll just remove the command.

src/main.rs
qsantos: Done

Cargo.toml
qsantos: Done

Cargo.toml
qsantos: Done

src/main.rs
qsantos: Done

tests/cmd/publish/mod.rs
qsantos: Done

qsantos:
> Are _all_ commands tested? That would be great because then we could finally close the issue that tracks that

Un-ignoring the test for package (assuming it was indeed due to git-lfs, which is now disabled), the following commands are not covered, according to `cargo-llvm-cov`:

- `uninstall`
- `generate`

Regarding generate, I guess it is indirectly tested when running `cargo build` in `sensor-server`, but that's in recompiled code that `cargo-llvm-cov` cannot see.

src/command.rs
tomkarw: Unless I'm mistaken, all of the test code is still compiled unconditionally (including release binary).

I meant not only the `test_registry`, but basically all of the code added as part of this PR.

qsantos: There is no more change to `src/`. All the code is now in `tests/`.

tomkarw (approved):
LGTM, great effort.

src/command.rs
tomkarw: True, I was looking at `Cargo.toml`, but missed that the deps now live in `dev-dependencies`. Resolving.

tests/test_registry.rs
```diff
@@ -0,0 +1,92 @@
...
+    body: Bytes,
+) {
+    tracing::info!("Uploaded package to {path} ({} bytes)", body.len());
+    state.write().unwrap().insert(path, body);
+}
+
+// do not use (flavor = "current_thread") here because the user-provided function is blocking
+#[tokio::main]
+pub async fn with_test_registry<F: FnOnce(&str)>(f: F) {
+    let port = PORT.fetch_add(1, Ordering::Relaxed);
```
tomkarw, line 66:
```
Port 0 can be used by applications when calling the bind() command to request the next available dynamically allocated source port number.
```

You could do SocketAddr::new([127, 0, 0, 1], 0) and get a different free port on each call. You can then call `SocketAddr::port` if you need to know which port was chosen.

mara-schulke (approved):

tests/cmd/tuto/in/sensor.proto
```diff
@@ -0,0 +1,25 @@
...
+  rpc ReadTemperature(DeviceId)
+    returns (Measurement);
+}
+
+// Device Identifier
+message DeviceId {
+  string id = 1;
+}
+
+// Temeperature measured by a device
```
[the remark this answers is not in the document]
qsantos: Good catch!

tests/test_registry.rs
qsantos: Oh, right! The code was initially starting a separate process, but now that it's simple `async` code, I can definitely do that.

Done!

tomkarw:
@qsantos Merge?

PR #201, opened 2023-12-15
Implement a canary edition system

### LINKED ISSUE #64: Encode manifest version in manifest
Yes - it's typical to store a 'manifest version' in the manifest so compatibility can be flagged if breaking changes are made. This avoids semantic misinterpretation, making it easy to ahead-of-time surface "buffrs is out of date" rather than surfacing "some field is missing".

e.g. store

```
manifest-version: "1.2"
```

and if buffrs reads a manifest it can't interpret (because it uses some incompatible feature) you can error that the manifest is out of date.

_Originally posted by @j-baker in https://github.com/helsing-ai/buffrs/pull/38#discussion_r1331132029_

mara-schulke:
This PR implements an edition system for buffrs manifests. This allows us to version manifests properly and show users errors (or map the manifests internally behind the scenes)

```
Error:   × could not deserialize Proto.toml
  ╰─▶ TOML parse error at line 1, column 1
        |
      1 | edition = "0.7"
        | ^^^^^^^^^^^^^^^
      unsupported manifest edition, supported editions of 0.8.0 are: 0.8
```

This is a preliminary feature to roll out before driving wide adoption of buffrs to be able to still make breaking changes (and introduce new features) while having an existing user base.

Closes #64 

---

## Canary Edition

The canary edition correlates (and de/encodes to `0.x` where `x` is the current minor version of buffrs). This means that until the stable release there is an option to introduce potentially breaking changes with every minor release.

## Unknown Edition

Every other manifest release edition that is not matching the current canary one is deserialized into `unknown` and thus triggers an error for users.

---

## Introduces

- `pub enum Edition` – The edition enum
- `pub const CANARY_EDITION` – The current canary edition of this crate

## Changes

- `Manifest` / `RawManifest`s internal layouts to accommodate the edition logic
- Implements a custom serializer / deserializer for branching and adding backwards compatibility
- The semantics of `buffrs package`: The compressed package will always contain a manifest with the current canary edition inside. Even if the "source" manifest didn't contain any edition pin
- The edition is optional in the `Porto.toml` and inferred to be the canary edition, which means that users are responsible for keeping it compatible (this also ensures that this feature works out of the box with existing projects!)

---

### Editioned (supported)

```toml
edition = "0.7"

[package]
name = "xyz"
version = "xyz"
type = "lib"
```

### Uneditioned (supported)

```toml
[package]
name = "xyz"
version = "xyz"
type = "lib"
```

### Unknow / Outdated (unsupported)

```toml
edition = "0.6"

[package]
name = "xyz"
version = "xyz"
type = "lib"
```

This causes the following error:

```
Error:   × could not deserialize Proto.toml
  ╰─▶ TOML parse error at line 1, column 1
        |
      1 | edition = "0.6"
        | ^^^^^^^^^^^^^^^
      unsupported manifest edition, supported editions of 0.8.0 are: 0.8
```

---

## Open Points

After the review of the core concept of editions the following points are open and must be addressed before merging:

- [x] Document this in the guide
- [x] Write tests for this


src/manifest.rs
```diff
@@ -33,15 +33,212 @@ use crate::{
...
+        where
+            D: Deserializer<'de>,
+        {
+            struct EditionVisitor;
+
+            impl<'de> serde::de::Visitor<'de> for EditionVisitor {
+                type Value = Edition;
+
+                fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+                    formatter.write_str("`Alpha` or `Beta`")
```
mara-schulke, line 165:
```suggestion
                    formatter.write_str("a valid edition string")
```

src/manifest.rs
```diff
@@ -33,15 +33,212 @@ use crate::{
...
+            deserializer.deserialize_str(EditionVisitor)
+        }
+    }
+
+    impl<'de> Deserialize<'de> for RawManifest {
+        fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+        where
+            D: Deserializer<'de>,
+        {
+            static FIELDS: &[&str] = &["package", "dependencies"];
```
mara-schulke, line 188: Design choice: Should this contain `edition`?

docs/src/reference/editions.md
```diff
@@ -0,0 +1,40 @@
...
+```toml
+edition = "0.7"
+```
+
+Canary editions are short-lived editions that are attached to a specific
+minor release of buffrs in the `0.x.x` version range. The edition name contains
+the minor version it is usable for. E.g. the edition `0.7` is usable /
+supported by all `0.7.x` buffrs releases. Compatibility beyond minor releases
+is not guaranteed as fundamental breaking changes may be introduced between
+editions.
```
tomkarw, line 40: Why the special case for `0.x` versions and calling them Canary? [SemVer](https://semver.org/#spec-item-4) explicitly allows breaking changes pre `1.0` release, we don't have to special-case it.

I understand we want a mechanism for differentiating between minor and braking changes, but that's what major and minor versions are for. We are redefining semantic versioning (`major -> minor, minor -> patch`), causing potential confusion, for the questionable gain of keeping the major version below `1`?

So we either release `1.0` today, and then can release non-breaking changes as `1.y` and breaking changes as `x.0`, or we stick with `0.y` and live with the fact that not every minor release breaks things, while each requires updating buffrs packages.

Finally, a nitpick, but I think the word `Canary` is misused here. My understanding of [canary release](https://martinfowler.com/bliki/CanaryRelease.html) is to deploy a new version for only a small subset of consumers while monitoring for potential issues. We are not doing that, instead introducing a new version, while not forcing it on users, and with no intention of reverting. May I propose `Minor Editions` or `Patch Editions`, as that's literally what they are?

src/manifest.rs
```diff
@@ -33,15 +33,212 @@ use crate::{
...
+
+/// Edition of the buffrs manifest
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Edition {
+    /// The canary edition of manifests
+    ///
+    /// This indicates that breaking changes and unstable behavior can occur
+    /// at any time. Users are responsible for consulting documentation and
+    /// help channels if errors occur.
+    Canary,
```
tomkarw, line 47: This special-casing is what I worry about. What changes will be needed here once we release `v1.0`?

src/manifest.rs
```diff
@@ -33,15 +33,212 @@ use crate::{
...
+            Self::Unknown { package, .. } => package.as_ref(),
+        }
+    }
+
+    fn dependencies(&self) -> &DependencyMap {
+        match self {
+            Self::Canary { dependencies, .. } => dependencies,
+            Self::Unknown { dependencies, .. } => dependencies,
+        }
+    }
```
tomkarw, line 90: Now all of this logic needs to be duplicated, and will have to be changed again for `v1`, but I can't understand it's justification.

src/manifest.rs
tomkarw: I believe it should.

docs/src/reference/editions.md
mara-schulke: So (ccing @xfbs) the meaning behind canary in the software comes from canary birds used to detect gas in coal mines, so translated this means "may be dangerous", hence canary editions of buffrs are intended to be used by early adopters that are ok with hitting issues.

Regarding the Edition system, please see the thread in slack with @Tpt - these editions are not meant to be supported longterm, but serve as a way to detect when / and with which buffrs version a package was released. This is independent from the semver meaning.

src/manifest.rs
mara-schulke: It is not possible to have a:

```
struct RawManifest {
  edition: Option<Edition>,
  package: Option<..
  ..
}
```

here as the editions may contain different fields / data.

src/manifest.rs
mara-schulke: `Stable` can then be explicitly be compatible with certain older editions that are bijective to the new structure

docs/src/reference/editions.md
xfbs: +1 from me for getting canaries involved into even more software engineering :smile:

xfbs:
I just want to add one thing here, which is a pattern that I have used before that I found rather useful for when you have versioned data structures and getting them to parse easily with serde. From glancing at the code in this PR, it looks like this strategy might be useful here as well. Some of this explanation includes stuff we are already doing, but I repeat it here for completness' sake.

The idea is that we use serde's support for internally tagged enums to get the decoding for free. To do this, we can have an `Edition` enum, and a `Manifest` struct (this one is internal, it can change at any time):

```rust
// could be auto-derived from VersionedManifest using strum, for better maintainability
#[derive(Serialize, Deserialize)]
#[serde(rename_all = "snake_case")]
pub enum Edition {
    V1,
    V2,
}

// internal representation of a manifest (we are free to change this, as it is now decouple from the
// representation on disk).
#[derive(Serialize, Deserialize)]
pub struct Manifest {
    edition: Edition,
    authors: Vec<String>,
    frobnicate: bool,
}
```

Now, when we actually parse the manifests, we use a different type called a `VersionedManifest` (this is raw data that we parse from the filesystem). Note that this is now an enumeration.

```rust
// external representation of a manifest (add new variants here for new editions)
#[derive(Serialize, Deserialize)]
#[serde(tag = "edition", rename_all = "snake_case")]
pub enum VersionedManifest {
    V1(ManifestV1),
    V2(ManifestV2),
}
```

To translate this: with the `tag = "edition"`, we are telling serde that in our TOML there is a field called `edition`, which it should parse first to figure out which variant it should match. We also tell it to `snake_vase` the variant names. So basically, the edition field can now be `v1` and `v2`. We are also able to override them with something else using `#[serde(tag = "something")]`.

The variants of this enumeration should obviously align with the `Edition` enum. In fact, the `Edition` enum can be automatically derived from this `VersionedManifest` using the `strum` crate, if desired.

Now, we can define these manifests. These are public-facing and may never change (part of the public API, if you wish):

```rust
// on-disk representations for manifests, one for every edition
#[derive(Serialize, Deserialize)]
pub struct ManifestV1 {
    // these fields can be public because they may never change
    pub author: String,
}

#[derive(Serialize, Deserialize)]
pub struct ManifestV2 {
    pub authors: Vec<String>,
    pub frobnicate: bool,
}
```

You can see that these two have different shapes — one has an *author* which is a single string, and one has a list of *authors*. This is fine — we can totally change the shape between versions. We have full flexibility to make any kind of breaking changes in between editions.

Finally, we write a bit of glue code to convert a `VersionedManifest` into our internal `Manifest` representation:

```rust
// this is where we translate the public-facing versioned manifest into our
// internal representation. we can change our internal representation at any point
// and change how we do this translation, as long as we don't change the semantics
impl Into<Manifest> for ManifestV1 {
    fn into(self) -> Manifest {
        Manifest {
            edition: Edition::V1,
            authors: vec![self.author],
            frobnicate: false, // fallback to default value
        }
    }
}

impl Into<Manifest> for ManifestV2 {
    fn into(self) -> Manifest {
        Manifest {
            edition: Edition::V2,
            authors: self.authors,
            frobnicate: self.frobnicate,
        }
    }
}

impl Into<Manifest> for VersionedManifest {
    fn into(self) -> Manifest {
        match self {
            VersionedManifest::V1(manifest) => manifest.into(),
            VersionedManifest::V2(manifest) => manifest.into(),
        }
    }
}
```

What is nice about this is that now, the two toml files parse properly:

```toml
edition = "v1"
author = "Snoop Doggy Dogg"
```

and this one as well:

```toml
edition = "v2"
authors = ["Snoop Doggy Dogg", "Dr. med. Dre"]
frobnicate = true
```

And yet, once we have parsed them both fully and turned them into a `Manifest`, they produce the same shape of data (but yet we still have that `edition` field, so that we can in the code have different behaviours depending on which edition you have chosen, if you really want):

```
Manifest {
    edition: Edition::V1,
    authors: vec!["Snoop Doggy Dogg"],
    frobnicate: false,
}
```

I'm a really big fan of how easy serde makes it to compose things like this. We could even have support for unknown editions, by writing something like this:

```rust
// the untagged makes it try to match the input as a Known variant first, and if it cannot, fall
// back to keeping it as a string. so "v1" => Known(V1), "v4" => Unknown("v4").
#[derive(Serialize, Deserialize)]
#[serde(untagged)]
pub struct RawEdition {
    Known(Edition),
    Unknown(String),
}
```

> Not that we would ever want that, probably, but the way this stuff composes makes it possible to mix strongly-typed stuff with a fallback to loosely-typed, dynamic data.

Maybe this is something we could implement here as well, it might save us some custom `serde` parsing code?

xfbs:
Another cool trick with serde (while I'm at it): suppose you have some data structure that is typed, but maybe in the future you might add fields to it. You don't care about those fields, but you also don't want to lose them. So what do you do?

This:

```rust
#[derive(Serialize, Deserialize)]
pub struct MyType {
    author: String,
    uuid: Uuid,
    tags: BTreeSet<Tag>,

    // anything else not present when parsing gets dumped into this
    #[serde(flatten)]
    other: BTreeMap<String, serde_json::Value>,
}
```

Serde is so cool!

mara-schulke:
@xfbs so, I agree, and I have used internally tagged serde enums before for these kinds of work. The problem here is that the `Canary` edition does not serialize to one specific edition, but to a different string on every buffrs version hence the custom serialization logic is needed.

The custom deserializers for the `RawManifest` enable us to treat the missing tag differently from how serde would behave by default. So to summarize this quickly: Its not possible to teach serde the current implementation's behavior through combinations of `rename` and `tag`. This is why I opted to implement this manually.

If you find a way of having untagged variants default to a specific tag + rename a variant dynamically let me know! Im open for suggestions but after carefully consulting the docs and lib of serde this is just not covered by the derive syntax.

xfbs:
@mara-schulke I see! As far as I understand, we can certainly have untagged variants default to a specific tag. However, you are right indeed in that we cannot rename a variant dynamically, this is tracked in https://github.com/serde-rs/serde/issues/450.

As far as I understand, the manually-implemented behaviour can be (almost) reimplemented like this:

```rust
#[derive(Serialize, Deserialize)]
#[serde(tag = "edition")]
pub enum TaggedManifest {
    #[serde(rename = "0.7")]
    Canary(CanaryManifest),
}

#[derive(Serialize, Deserialize)]
#[serde(untagged)]
pub enum Manifest {
    Tagged(TaggedManifest),
    Untagged(FallbackManifest),
}
```

The only caveat being that due to the missing dynamic rename functionality, we *have* to write `rename = "0.7"` and we currently cannot write `rename = concat!("0.", env!("CARGO_MANIFEST_VERSION_MINOR"))`. So, if we are okay with having a place in the code that we need to manually update with the current version name (this can be unit tested quite easily), then this approach might be useful to us here.

If we don't want to have to manually update that `0.7`, then obviously implementing `Serialize` and `Deserialize` manually (as is the current approach) is the only way to go 😢

mara-schulke:
Well, but the current behavior of "untagged" manifests is to deserialize as the current canary edition. I don't think that this is possible as of right now?

xfbs:
So we only have two parsing cases:

- `edition = "0.7"` or *missing*: treat as current canary
- `edition = "0.6"` or any other, non-current edition: emit warning

Is that correct?

mara-schulke:
Well, no, because 0.7 may be compatible with 0.6, then we would not emit an error.

mara-schulke:
@xfbs @tomkarw @Tpt @qsantos: this is blocking the migration from kicking of, may we either merge this or get concrete change requests?👀🙈 

I'm open on reiterating the implementation, I'm just waiting on suggestions / concrete feedback here. If you don't have any, may we shift the discussion to a github discussion / issue and move on?

docs/src/reference/editions.md
```diff
@@ -0,0 +1,40 @@
...
+The edition system exists to allow for fast development of buffrs while
+allowing you to already migrate existing protobufs to buffrs even though it
+has not yet reached a stable version.
+
+Editions can be either explicitly stated in the `Proto.toml` or are
+automatically inlined once a package is created using buffrs. This ensures that
+you dont need to care about them as a user but get the benefits.
+
+> Note: If you release a package with an edition that is incompatible with
+> another one (e.g. `0.7` is incompatible with `0.8`) you will need to
```
Tpt, line 13: I would write "(e.g. *if* `0.7` is incompatible with `0.8`)" here to highlight that `0.8` might be compatible with `0.7` but it's not guaranteed.

src/manifest.rs
Tpt: Idea: I instead of `Stable` what about something like `1` to be able to handle a possible future edition `2` in a consistant manner. Am I correct?

In this line of thought what about renaming `Canary` to `Canary1`. Then if we introduce breaking change we would introduce `Canary2` and might just drop `Canary1` or keep compatiblity for some time with it?

src/manifest.rs
```diff
@@ -33,15 +33,212 @@ use crate::{
...
+    /// help channels if errors occur.
+    Canary,
+    /// Unknown edition of manifests
+    ///
+    /// This is unrecommended as breaking changes could be introduced due to being
+    /// in the beta release channel
+    Unknown,
+}
+
+impl Default for Edition {
```
Tpt, line 55: Nit: I would maybe avoid this to enforce concious choice of edition by callers. What about `Edition::latest()`

src/manifest.rs
```diff
@@ -33,15 +33,212 @@ use crate::{
...
-    package: Option<PackageManifest>,
-    #[serde(default)]
-    dependencies: DependencyMap,
+#[derive(Debug, Clone, PartialEq, Eq)]
+enum RawManifest {
+    Canary {
+        package: Option<PackageManifest>,
+        dependencies: DependencyMap,
+    },
+    Unknown {
```
Tpt, line 71: Genuine question: is `Unknown` useful if you are raising an error when parsing a manifest without edition?

src/manifest.rs
```diff
@@ -33,15 +33,212 @@ use crate::{
...
+                    formatter.write_str("a valid edition string")
+                }
+
+                fn visit_str<E>(self, value: &str) -> Result<Self::Value, E>
+                where
+                    E: serde::de::Error,
+                {
+                    match value {
+                        c if c == CANARY_EDITION => Ok(Edition::Canary),
+                        _ => Ok(Edition::Unknown),
```
Tpt, line 174: Idea: fail here with an unknown edition error to avoid having an `Unknown` variant in the `Edition` enum?

Tpt (approved):
The current state seems a great improvement to me. But I am not one of the project maintainer so it's not my call to make I think.

src/manifest.rs
mara-schulke: Yes ofc, the internal edition name (eg stable, or V1) would probably be serialized as edition = "1"

src/manifest.rs
mara-schulke: This is worth a revisit!

xfbs:
Sorry from my side for being slow!

- I think the code looks good, I also approve this
- I have a bit of uncertainty/lack of understanding of the strategy around the canary editions, backwards compatibility and all that good stuff
- Are we maybe overloading the edition key with the version, and what we really want is:

    ```toml
    edition = "alpha"
    buffrs_version = "0.7" # requires 0.7 at minimum
    ```
    
    and
    
    ```toml
    edition = "1"
    ```
    
    Or something like that (aka have separate fields for edition versioning and buffrs minimum versions)? (not sure if we do, just thinking out aloud)

- It scares me a little bit that the current approach is that you need to re-publish a package to upgrade to a new edition (and all of the dependencies need to do the same), if I understood that correctly (did I understand that correctly?)
- I feel like there are some thoughts that could be better discussed synchronously, because I feel bad when I ask stupid questions or write a wall of text :smile:

I think @Tpt left some good comments, and maybe I'm just overthinking it!

xfbs (approved):

docs/src/reference/editions.md
tomkarw:
```
Canary release is a technique to reduce the risk of introducing a new software version in production by slowly rolling out the change to a small subset of users before rolling it out to the entire infrastructure and making it available to everybody. 
```

I still don't think we're using the word in the right context here and I'm worried the meaning-overloading will be confusing to users familiar with the above definition. Still, this is a nit, so feel free to resolve this thread.

mara-schulke:
> It scares me a little bit that the current approach is that you need to re-publish a package to upgrade to a new edition (and all of the dependencies need to do the same), if I understood that correctly (did I understand that correctly?)

I think what you need to acknowledge is that you would need to do this either way. If we break the compat (for whatever reason) you can't use old packages. This is already the case as of right now and will always be the case prior to being stable. What this edition system solves is that you gain stability and awareness of when this is the case (and if it is the case, you get the ability to repair it, by mapping the old manifest to the new).

Re mixing editions and versions: buffrs doesn't have editions as rust has them. We only version the manifest through editions thus there is no need to use "alpha", "beta" etc for the manifest version. The only version we track is the stability of the package manager and thus strictly related to the server version string of it. The versioning system described by this pr is basically just formalizing what is already the case: every new minor version (pre 1.0) _may_ be incompatible with the prior one. The manifest may change completely between each version and thus the edition tracks when the manifest was valid / how it should look like and enables us to write migration guides between each edition.

Most likely each breaking change in buffrs (and thus each new minor version) will come along with some, potentially breaking, manifest change. If that is not the case anymore we are approaching stable.

docs/src/reference/editions.md
mara-schulke: Just as a hint: the canary system is internal and not necessarily user facing, also editions are automated and not necessarily managed by the user itself.

xfbs:
You are obviously much more knowledgeable on buffrs than me, so always take what I say with a grain of salt 😅 

> If we break the compat (for whatever reason) you can't use old packages. This is already the case as of right now and will always be the case prior to being stable. 

We *could* make a conscious choice to say: *while buffrs is unstable (there will be breaking changes in it's API), we are explicitly versioning the manifests and guaranteeing that newer versions will be able to use older manifests*. if we wanted to, anyways. 

My understanding is that that is not something we want to do, because we want to be able to make many breaking changes or changes to the manifests that are not easily "convertible" (aka bijective)?

> Most likely each breaking change in buffrs (and thus each new minor version) will come along with some, potentially breaking, manifest change. If that is not the case anymore we are approaching stable.

Does the manifest change often? I think I was under the impression that it is relatively stable — but I am likely wrong!

Sorry for opening a can of worms with my comment but I think this explanation from you really helped me understand the motivation behind the design the way it is 😊

src/manifest.rs
mara-schulke: Re: Yes.

This is used to omit a re-serialization of the manifest that users use (to make the edition management a virtual thing)

Example:

The manifest that a user hast may be:

```
[package]
type = "lib"
name = "xyz"
version = "1.0.0"
```

which _could_ be (or in fact is) a valid manifest across (or since) many different editions as these keys are enough to publish a lib without any deps.

If we would not deserialize this manifest as "unknown" but as "canary" / "0.7" – we would set the edition after every buffrs command that edits the manifest.

This would resolve in an actual incompatibility (e.g. if 0.7 is incompatible with 0.8 due to a field being added in the dependencies section) even though it is actually compatible / deserializable with the 0.8 schema.

E.g.

```
edition = "0.7"

[package]
type = "lib"
name = "xyz"
version = "1.0.0"
```

would yield an error in case buffrs 0.8 is used in this  project and the user would have to manually update the manifest to be 0.8 here, but there is no migration to be done:

```
edition = "0.8"

[package]
type = "lib"
name = "xyz"
version = "1.0.0"
```

instead when using:

```
[package]
type = "lib"
name = "xyz"
version = "1.0.0"
```

the edition is inferred / treated to be the latest (but internally stored as the `Unknown` varianted and _only_ serialized once a package is being created for the publishing.

This results in the behavior of automatic inlining once `buffrs publish` is called:

(content of `Proto.toml`):

```
[package]
type = "lib"
name = "xyz"
version = "1.0.0"
```

(content of `Proto.toml` in the `lib-1.0.0.tgz`):

```
edition = "0.7"

[package]
type = "lib"
name = "xyz"
version = "1.0.0"
```

mara-schulke: I hope this was somewhat clear, lmk if I explained it wrong (or in an messy way)

src/manifest.rs
mara-schulke: See above! Im happy to convert this behavior to a GitHub discussion where we can debate about the longterm management of this?

PR #202, opened 2023-12-18
Make buffrs compatible with legacy packages

qsantos:
- `vendor/package_name/defs.proto` → `vendor/defs.proto`
- `import "package_name/defs.proto"` → `package "package-name/defs.proto"`
- `package module_name` → `package clarity.module_name`
- `module_name.type x;` → `clarity module_name.type x;`


tomkarw:
Hey @qsantos, can you add a quick post-mortem/rationale as to why this was closed? I'm sure there was a good reason, but I'd like it to be preserved in writing.

Issue #192, opened 2023-12-08
`buffrs install` fails when the `proto/vendor/`directory does not exist


qsantos:
```

qsantos:
Putting aside #192, it looks like the intended behavior is indeed to fail to run when `proto/` does not exist:

https://github.com/helsing-ai/buffrs/blob/4a6d0e9084638f40909052aa1894bc20facbc01b/src/package/store.rs#L58-L67

So the question behind this issue is now: should the `proto/` directory be created automatically?

mara-schulke:
Yes, but in `install` we should check for existence and initialize it if it doesn't exist.

PR #203, opened 2023-12-18
Create `proto` and `vendor` directories when running commands

ntrinquier:
Fixes https://github.com/helsing-ai/buffrs/issues/192

### DIFF
--- a/src/package/store.rs
+++ b/src/package/store.rs
@@ -48,23 +48,12 @@ impl PackageStore {
         Self::open(&current_dir().into_diagnostic()?).await
     }
 
-    /// Check if this store exists
-    async fn exists(&self) -> miette::Result<bool> {
-        let meta = fs::metadata(&self.proto_path()).await.into_diagnostic()?;
-
-        Ok(meta.is_dir())
-    }
-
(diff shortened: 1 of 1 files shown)


src/command.rs
```diff
@@ -252,6 +252,8 @@ pub async fn install() -> miette::Result<()> {
     let credentials = Credentials::load().await?;
 
     store.clear().await?;
+    PackageStore::create(store.proto_path()).await?;
```
mara-schulke, line 255: This should be done when calling `open` to prevent the issue across all commands

src/package/store.rs
```diff
@@ -59,6 +59,9 @@ impl PackageStore {
     async fn open(path: &Path) -> miette::Result<Self> {
         let store = Self::new(path.into());
 
+        PackageStore::create(store.proto_path()).await?;
+        PackageStore::create(store.proto_vendor_path()).await?;
+
         if !store.exists().await? {
```
mara-schulke, line 65: Just as a note, this is now useless

mara-schulke (approved):

Issue #45, opened 2023-09-12
Setup a CI for the registry + Dockerize the binary


mara-schulke:
The binary doesnt do anything as the registry is not implemented but having the overall setup would be great!

PR #206, opened 2024-01-05
Add Dockerfile for registry

BlowaXD:
This should partially solve #45 

## Additions
- added `registry/Dockerfile`
- added `registry/.dockerignore` ignoring `**/target/**` to avoid having to sending too much files to context
- added README.md in registry project with docker build command

I decided to manually install protobuff deps to have an easy way to manage the protobuf compiler version


tomkarw:
Shouldn't the `Dockerfile` live in `/registry` folder?

mara-schulke:
Hei @BlowaXD, tysm for your contribution! 😊

BlowaXD:
> I agree with @tomkarw's comment, otherwise this looks good to me 😊

Sure, I did first assume that someone would have wanted to build from the same directory

Moving it to `registry` folder requires either:

Having this `COPY` statement instead of the current one
```dockerfile
COPY .. .
```
or using this command to build
```sh
docker build -t registry . --file registry/Dockerfile
```

I don't know which UX you want to provide, I do not see an alternative because of the registry requiring `buffrs` as a dependency

tomkarw:
> I don't know which UX you want to provide, I do not see an alternative because of the registry requiring `buffrs` as a dependency

I'd go with `COPY .. .`. 

I think the issue is more general, we have a CLI, a registry, and they share some common parts. This clearly indicates 3 crates (`cli`, `registry` and `buffrs-lib`), but currently `cli.contains(buffrs-lib)`. I'd move the Dockerfile and hack it to work, and we can open a ticket about at some point extracting the common parts out.

tomkarw (approved):

registry/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+    unzip protoc-${PB_VERSION}-linux-x86_64.zip -d ./proto
+
+RUN chmod 755 -R ./proto/bin
+ENV BASE=/usr
+RUN cp ./proto/bin/protoc ${BASE}/bin/
+RUN cp -R ./proto/include/* ${BASE}/include/
+
+WORKDIR /usr/src/buffrs
+RUN cd /usr/src/buffrs
+COPY .. .
```
NiklasRosenstein, line 20: You can't copy something from outside the build context.

Did a small test and it seems `COPY .. .` behaves the same as `COPY . .`.

BlowaXD:
Edited the PR, I gave a command in the `registry/README` to guide user to build the docker so it will be ready when `registry` works
As `registry` is a WIP, I didn't know what to write in the README so I decided to add a short disclaimer that can be removed whenever

mara-schulke:
Awesome, thank you @BlowaXD!

PR #210, opened 2024-01-09
add registry publish endpoint basis

BlowaXD:
WIP

This PR intends to add a working publish endpoint with a basic in memory metadata storage (that can be further implemented as a proper storage with #173)

### DIFF
--- a/.github/workflows/registry.yml
+++ b/.github/workflows/registry.yml
@@ -36,14 +36,6 @@ jobs:
         with:
           compose-file: "./registry/docker-compose.yml"
       - run: rustup update
-      # run registry locally to permit API tests
-      - name: "Run Registry in background"
-        working-directory: registry
-        run: |
-          cargo build --release
-          ./target/release/buffrs-registry --filesystem-storage /tmp &
-        env:
-          LISTEN: 127.0.0.1:4367
       - uses: Swatinem/rust-cache@v2
       - run: cd registry && cargo test --workspace
         env:
@@ -72,14 +64,6 @@ jobs:
       - run: rustup component add llvm-tools-preview
       - uses: Swatinem/rust-cache@v2
       - run: cargo install cargo-llvm-cov || true
-      # run registry locally to permit API tests
-      - name: "Run Registry in background"
-        working-directory: registry
-        run: |
-          cargo build --release
-          ./target/release/buffrs-registry --filesystem-storage /tmp &
-        env:
-          LISTEN: 127.0.0.1:4367
       - run: cd registry && cargo llvm-cov --workspace --fail-under-lines "$MINIMUM_LINE_COVERAGE_PERCENT"
 
 # TODO: This already runs across registiry folder in cli.yml, maybe it's time to extract common.yml?
--- a/.github/workflows/registry.yml
+++ b/.github/workflows/registry.yml
@@ -36,6 +36,14 @@ jobs:
         with:
           compose-file: "./registry/docker-compose.yml"
       - run: rustup update
+      # run registry locally to permit API tests
+      - name: "Run Registry in background"
+        working-directory: registry
+        run: |
+          cargo build --release
+          ./target/release/buffrs-registry --filesystem-storage /tmp &
+        env:
+          LISTEN: 127.0.0.1:4367
       - uses: Swatinem/rust-cache@v2
       - run: cd registry && cargo test --workspace
         env:
@@ -64,6 +72,14 @@ jobs:
       - run: rustup component add llvm-tools-preview
       - uses: Swatinem/rust-cache@v2
       - run: cargo install cargo-llvm-cov || true
+      # run registry locally to permit API tests
+      - name: "Run Registry in background"
+        working-directory: registry
+        run: |
+          cargo build --release
+          ./target/release/buffrs-registry --filesystem-storage /tmp &
+        env:
+          LISTEN: 127.0.0.1:4367
       - run: cd registry && cargo llvm-cov --workspace --fail-under-lines "$MINIMUM_LINE_COVERAGE_PERCENT"
 
 # TODO: This already runs across registiry folder in cli.yml, maybe it's time to extract common.yml?
--- a/registry/src/api.rs
+++ b/registry/src/api.rs
@@ -15,12 +15,24 @@
 //! # Buffrs Registry API
 
 use crate::context::Context;
+use crate::proto::buffrs::registry::registry_server::RegistryServer;
+use eyre::Result;
+use tonic::transport::Server;
 
 mod grpc;
 
 impl Context {
     /// Launch API
-    pub async fn launch_api(&self) {
-        todo!()
+    pub async fn launch_api(&self) -> Result<()> {
+        let ctx = self.clone();
+
+        let server = Server::builder().add_service(RegistryServer::new(ctx));
+
+        let listen_address = self.listen_address();
+        tracing::info!("Starting server: {:?}", listen_address);
+
+        server.serve(listen_address).await?;
+
+        Ok(())
     }
 }
--- a/registry/src/api/grpc.rs
+++ b/registry/src/api/grpc.rs
@@ -12,36 +12,144 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::str::FromStr;
+
 use crate::{
     context::Context,
+    metadata::MetadataStorageError,
+    proto::buffrs::package::Type,
     proto::buffrs::registry::{
         registry_server::Registry, DownloadRequest, DownloadResponse, PublishRequest,
         PublishResponse, VersionsRequest, VersionsResponse,
     },
+    types::PackageVersion,
 };
 use async_trait::async_trait;
-use tonic::{Request, Response, Status};
+use buffrs::{manifest::PackageManifest, package::PackageName, package::PackageType};
+use semver::Version;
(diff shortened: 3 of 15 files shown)


registry/src/context.rs
```diff
@@ -18,23 +18,31 @@
...
     storage: AnyStorage,
+
+    listen_address: SocketAddr,
 }
 
 impl Context {
     /// Create a new context from a metadata instance and a storage instance.
-    pub fn new(storage: AnyStorage) -> Self {
-        Self { storage }
+    pub fn new(storage: AnyStorage, options: SocketAddr) -> Self {
```
mara-schulke, line 35: what's `options`?

registry/src/api.rs
```diff
@@ -15,12 +15,25 @@
...
 
 impl Context {
     /// Launch API
-    pub async fn launch_api(&self) {
-        todo!()
+    pub async fn launch_api(&self) -> Result<()>  {
+        let ctx = self.clone();
+
+        let server = Server::builder()
+        .add_service(RegistryServer::new(ctx));
```
mara-schulke, line 30:
```suggestion
            .add_service(RegistryServer::new(ctx));
```

(`cargo fmt` should be able to fix this :))

registry/src/api/grpc.rs
```diff
@@ -12,36 +12,110 @@
...
+        request: Request<PublishRequest>,
     ) -> Result<Response<PublishResponse>, Status> {
-        todo!()
+        let req: PublishRequest = request.into_inner();
+
+        // this shouldn't happen as package isn't optional
+        let package: Compressed = match req.package {
+            None => { return Err(Status::invalid_argument("package wasn't set")); },
+            Some(_value) => {_value}
+        };
```
mara-schulke, line 48: This looks good!

You can rewrite this a bit more condensed:

```rust
let Some(package) = req.package else {
    return Err(Status::invalid_argument("package wasn't set")); }
};
```

mara-schulke: Or even:

```rust
let package = req.package.ok_or(Status::invalid_argument("package wasn't set"))?;
```

registry/src/api/grpc.rs
```diff
@@ -12,36 +12,110 @@
...
+
+        // this shouldn't happen as metadata isn't optional
+        let metadata = match package.metadata {
+            None => { return Err(Status::invalid_argument("metadata wasn't set")); },
+            Some(_value) => {_value} 
+        };
```
mara-schulke, line 54: Same applies here :)

match statements are generally best when you need to branch a lot, if you just want to unpack something there are easier ways :)

registry/src/api/grpc.rs
```diff
@@ -12,36 +12,110 @@
...
+
+        // need to add some more checks (version conflict, do we allow overriding?)
+
+        let storage = self.storage();
+
+        let package_name = PackageName::from_str(metadata.name.as_str());
+        if let Err(error) = package_name {
+            // todo add some logging & proper error handling
+            return Err(Status::invalid_argument(format!("Package name isn't correct, {:?}", error)));
+        }
```
mara-schulke, line 77: You can use `.map_err` here :)

```
let package_name = PackageName::from_str(metadata.name.as_str())
   .map_err(|_| Status::invalid_argument(format!("Package name isn't correct, {:?}", error)))?;
```

registry/src/api/grpc.rs
```diff
@@ -12,36 +12,110 @@
...
+
+        let package_version = Version::from_str(metadata.version.as_str());
+        if let Err(_error) = package_name {
+            // todo add some logging & proper error handling
+            return Err(Status::invalid_argument("version isn't correct"));
+        }
+
+        let package_version = &PackageVersion{ package: package_name.unwrap(), version: package_version.unwrap() };
```
mara-schulke, line 85: `.unwrap` is bad practice generally

(if you use the result functions as above, its not needed anymore :))

registry/src/api/grpc.rs
```diff
@@ -12,36 +12,110 @@
...
+
+        let vec_ref = package.tgz.as_ref();
+        let package_bytes: &[u8] = vec_ref;
+
+        let storage_resp = storage.package_put(package_version, package_bytes).await;
```
mara-schulke, line 90:
```suggestion
storage.package_put(package_version, &package.tgz)
    .await
    .map_err(|_| Status...)?;
```

registry/src/api/grpc.rs
```diff
@@ -12,36 +12,110 @@
...
+        if let Err(_error) = storage_resp {
+            // todo add some logging & resiliency
+            return Err(Status::internal("Something went wrong on our side, sorry :("))
+        }
+
+        // todo: add to DB
+        // on failure, delete from storage?
+
+        return Ok(Response::new(PublishResponse{
```
mara-schulke, line 99:
```suggestion
        Ok(Response::new(PublishResponse{}))
```

.github/workflows/registry.yml
```diff
@@ -36,6 +36,14 @@ jobs:
         with:
           compose-file: "./registry/docker-compose.yml"
       - run: rustup update
+      # run registry locally to permit API tests
+      - name: "Run Registry in background"
```
mara-schulke, line 40: We definitely should not do this imho!

We can run the registry virtually through cargo test :) 

If you need help with this we can pair here 😊

registry/src/metadata.rs
```diff
@@ -0,0 +1,81 @@
...
+    #[error("Internal Error")]
+    InternalStuff,
+}
+
+/// [`MetadataStorage`] Instance
+pub type AnyMetadataStorage = Arc<dyn MetadataStorage>;
+
+/// inmemory storage
+///
+pub fn new_storage() -> AnyMetadataStorage {
```
mara-schulke, line 65: I would drop this function tbh

registry/src/metadata/memory.rs
```diff
@@ -0,0 +1,95 @@
...
+        let version_string = package.version.to_string();
+
+        let Some(versions_mutex) = packages.get(name_string.as_str()) else {
+            return Err(MetadataStorageError::PackageMissing(
+                name_string,
+                version_string,
+            ));
+        };
+
+        let versions = versions_mutex.lock().unwrap();
```
mara-schulke, line 51: Unwrapping is never a good idea!

So you could either propagate the `Err` or you can use tokios mutex impl, which lets you await the lock (im 99% confident that you don't need to handle errors in that case)

registry/src/metadata/memory.rs
```diff
@@ -0,0 +1,95 @@
...
+        let Some(versions_mutex) = packages.get(name_string.as_str()) else {
+            let mut versions_hashmap = HashMap::default();
+            let _ = &versions_hashmap.insert(version_string, package);
+            let _ = &packages.insert(name_string, Mutex::new(versions_hashmap));
+
+            return Ok(());
+        };
+
+        let Ok(mut versions) = versions_mutex.lock() else {
+            return Err(MetadataStorageError::InternalStuff);
```
mara-schulke, line 81: Same here

registry/src/metadata.rs
```diff
@@ -0,0 +1,81 @@
...
+    #[error("duplicate package")]
+    PackageDuplicate(String, String),
+
+    /// Unknown error
+    #[error(transparent)]
+    Other(#[from] SharedError),
+
+    /// Unknown error
+    #[error("Internal Error")]
+    InternalStuff,
```
mara-schulke, line 57:
```suggestion
    Internal,
```

src/package/type.rs
```diff
@@ -26,6 +26,18 @@ pub enum PackageType {
     Api,
 }
 
+impl TryFrom<i32> for PackageType {
```
mara-schulke, line 29: If you want to use this for protobuf handling, we should definitely impl 

`From<crate::proto::buffrs_registry::package::PackageType> for PackageType`

You can then go from `i32 -> crate::proto::buffrs_registry::package::PackageType -> PackageType`

The reason im flagging this is because if the proto file changes, this impl will stay the same which might lead to bugs. If we impl type conversion on the generated code stubs we don't convert explicitly but let tonic generate this code based on the proto file

registry/src/api/grpc.rs
```diff
@@ -110,10 +110,9 @@ impl Registry for Context {
             .await
             .map_err(|_| Status::internal("Something went wrong on our side, sorry :("))?;
 
-        let package_type_raw_value = metadata.r#type;
-        let package_type = PackageType::try_from(package_type_raw_value).map_err(|_| {
-            Status::invalid_argument(format!("Unknown Package Type: {}", package_type_raw_value))
-        })?;
+        let package_type_raw_value = crate::proto::buffrs::package::Type::try_from(metadata.r#type)
+            .expect("Shouldn't happen");
```
mara-schulke, line 114: Also don't use expect! (Except in tests)

In functions that return a `Result` you can always use `.map_err` or `.wrap_err` (from eyre / miette) to map this to an error type that you can propagate with `?`

registry/src/api/grpc.rs
```diff
@@ -110,10 +110,9 @@ impl Registry for Context {
...
+        let package_type = PackageType::from(package_type_raw_value);
```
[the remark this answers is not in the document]
mara-schulke: (would be a more functional way)

mara-schulke (approved):

PR #212, opened 2024-01-11
Cleanup service testing to use tokio channel and run client/server connection

BlowaXD:


registry/src/metadata/memory.rs
```diff
@@ -34,40 +28,44 @@ impl MetadataStorage for InMemoryMetadataStorage {
...
             .map_err(|_| MetadataStorageError::Internal)?;
 
-        let Some(package) = versions.get(version_string.as_str()) else {
-            return Err(MetadataStorageError::PackageMissing(
-                name_string,
-                version_string,
-            ));
-        };
+        let package =
+            versions
```
mara-schulke, line 52: this should be formatted differently by cargo fmt?

registry/tests/api/main.rs
```diff
@@ -23,30 +31,65 @@ pub fn create_publish_request_sample() -> PublishRequest {
...
-    let socket = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 0)), 1111);
-    let fake_server = Context::new(storage, metadata, socket);
+
+    let url = "0.0.0.0:0";
+
+    // this is needs to be removed once Context got cleaned up
+    let socket = SocketAddr::from_str(url).expect("Shouldn't happen");
+    let context = Context::new(storage, metadata, socket);
+
+    let _handle = tokio::spawn(async move {
```
mara-schulke, line 47:
```suggestion
    tokio::spawn(async move {
```

registry/tests/api/main.rs
```diff
@@ -23,30 +31,65 @@ pub fn create_publish_request_sample() -> PublishRequest {
     }
 }
 
-#[tokio::test]
-async fn test_publish_registry() -> Result<(), Box<dyn std::error::Error>> {
+pub async fn basic_setup() -> Result<RegistryClient<Channel>, Box<dyn std::error::Error>> {
```
mara-schulke, line 34: Two things here:

1. Box<dyn std::error::Error>> is rarely a good idea to use directly in your code base. There are two options: a) use `eyre`/`miette` or b) use concrete error enums to cover for all your cases. I think eyre is the most appropriate here
2. Stuff like this (which you want to "paste" into many functions") could/would be a macro ideally, this also frees you from certain drop restrictions (making it a function works nonetheless, its more of a nit/style question)

mara-schulke: PS: as we are in tests, its ok to use `unwrap` or `expect` to skip error handling entirely, this would simplify your types a lot

registry/tests/api/main.rs
```diff
@@ -23,30 +31,65 @@ pub fn create_publish_request_sample() -> PublishRequest {
...
     }
 
     // 2. Insert the same package for a duplicate check
     {
         // duplicate check
         let req = tonic::Request::new(create_publish_request_sample());
-        let Err(res) = fake_server
-            .publish(req)
-            .await
-            .map_err(|status| status.code())
```
mara-schulke, line 48: Sorry for advising you initially to do the map_err thing, I didn't recognize we are in the integration tests

registry/tests/api/main.rs
```diff
@@ -23,35 +29,68 @@ pub fn create_publish_request_sample() -> PublishRequest {
...
+
+#[tokio::test]
+async fn test_publish_registry() {
+    let mut client = basic_setup().await;
 
     // 1. Insert a package and expect it to be successful
     {
         let req = tonic::Request::new(create_publish_request_sample());
-        let _res = fake_server.publish(req).await?;
+        client.publish(req).await.expect("Shouldn't happen");
```
mara-schulke, line 84: nit: if you don't provide any meaningful content in the expect statement id argue that you can just use unwrap

mara-schulke (approved):
This looks great!!

PR #215, opened 2024-01-16
Add PgSQL metadata storage

BlowaXD:
- docker compose will now automatically apply migrations
- added a collate to facilitate semantic versions search

- [ ] fetching versions
- [ ] test fetching versions

- [ ] publishing version
- [ ] test publishing version


registry/src/metadata/postgresql.rs
```diff
@@ -0,0 +1,349 @@
...
+                self.name.to_string(),
+                self.version.to_string(),
+            ));
+        }
+        Ok(())
+    }
+}
+
+impl PgsqlMetadataStorage {
+    /// try from
```
tomkarw, line 188: these seem like a placeholder, but this MR is a Draft, so no biggie

registry/src/model/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#![allow(missing_docs)]
+
+use atmosphere::prelude::*;
```
tomkarw, line 16: I personally dislike both star imports (they obfuscate what's actually used and where it's pulled from) and preludes.

registry/tests/api/main.rs
```diff
@@ -36,12 +50,16 @@ pub fn create_list_versions_request_sample(version: String) -> VersionsRequest {
...
 
-pub async fn basic_setup() -> RegistryClient<Channel> {
+pub async fn basic_setup(pool: PgPool) -> RegistryClient<Channel> {
     let (client, server) = tokio::io::duplex(1024);
 
     let path = Path::new("/tmp");
     let storage = Arc::new(storage::Filesystem::new(path));
-    let metadata = Arc::new(InMemoryMetadataStorage::new());
+
+    let pgsql = buffrs_registry::metadata::postgresql::PgsqlMetadataStorage::new(pool);
```
tomkarw, line 59: Why the qualification instead of `use` statement?

src/package/type.rs
```diff
@@ -25,6 +25,16 @@ pub enum PackageType {
     Api,
 }
 
+impl From<i32> for PackageType {
+    fn from(value: i32) -> Self {
+        match value {
+            1 => PackageType::Lib,
+            2 => PackageType::Api,
+            _ => panic!("Wrong value"),
+        }
```
mara-schulke, line 34: if you can convert types, you need `TryFrom` fyi

registry/src/metadata/postgresql.rs
```diff
@@ -160,36 +162,40 @@ impl Publish<PgsqlMetadataStorage> for PackageManifest {
...
                 MetadataStorageError::Internal
             })?;
 
-        // insert failed, todo proper error handling
-        if pg_result.rows_affected() != 1 {
-            return Err(MetadataStorageError::PackageDuplicate(
-                self.name.to_string(),
-                self.version.to_string(),
-            ));
+        let commit_result = tx.commit().await;
```
tomkarw, line 183: Not quite right. Now you're doing exactly one query as part of the transaction, making it basically unnecessary. I line [140](https://github.com/helsing-ai/buffrs/pull/215/commits/05db820f49f395d60dc6c0be9abfc19d5ab59672#diff-b005606de71552e393b447fc882ecd3d9cfeedf4057e914209cf536cccc60507R140) you do a potential insert.

If `INSERT_VERSION_QUERY` fails, should we rollback `SELECT_OR_INSERT_PACKAGE_QUERY`? If yes, the transaction need to start before the latter and include both queries, if not, you can just remove the transaction all-together and go back to how this code looked before.

registry/src/metadata.rs
```diff
@@ -52,31 +53,37 @@ pub enum MetadataStorageError {
...
+}
 
-    /// Fetches a package with a given version and all packages above that version
-    async fn get_versions(
-        &self,
-        package_name: PackageName,
-        version: Option<VersionReq>,
+/// Fetching data storage
+#[async_trait::async_trait]
+pub trait FetchAllMatching<Executor> {
```
cdbrkfxrpt, line 74:
```suggestion
pub trait FetchMatching<Executor> {
```

registry/src/metadata/memory.rs
```diff
@@ -59,13 +57,16 @@ impl MetadataStorage for InMemoryMetadataStorage {
 
         Ok(package.clone())
     }
+}
 
-    async fn get_versions(
-        &self,
+#[async_trait::async_trait]
+impl FetchAllMatching<InMemoryMetadataStorage> for PackageManifest {
```
cdbrkfxrpt, line 63:
```suggestion
impl FetchMatching<InMemoryMetadataStorage> for PackageManifest {
```

registry/src/metadata/postgresql.rs
```diff
@@ -0,0 +1,319 @@
...
+                Some(package.version.to_string()),
+            ));
+        };
+
+        Ok(PackageManifest::try_from(searched)?)
+    }
+}
+
+#[async_trait::async_trait]
+impl FetchAllMatching<PgsqlMetadataStorage> for PackageManifest {
```
cdbrkfxrpt, line 59:
```suggestion
impl FetchMatching<PgsqlMetadataStorage> for PackageManifest {
```

cdbrkfxrpt (changes requested):

registry/migrations/20230828111733_users.up.sql
```diff
@@ -1,37 +1,36 @@
 -- table of users.
-CREATE TABLE "users" (
-    "id"          SERIAL PRIMARY KEY,
-    "handle"      TEXT NOT NULL UNIQUE,
-    "created_at"  TIMESTAMPTZ NOT NULL DEFAULT (now()),
-    "updated_at"  TIMESTAMPTZ NOT NULL DEFAULT (now()),
+CREATE TABLE users (
```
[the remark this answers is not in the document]
tomkarw: I meant that more as a general principle. The idea of having migrations in the first place is that it's an append-only log. If we're in a stage where we can modify existing migrations, we could as well squash them and operate on just one *.sql file.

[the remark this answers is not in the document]
NiklasRosenstein: You should squash all migrations introduced during the development cycle of a PR, or otherwise you accumulate tons of unnecessary migration steps.

I don't know about sqlx, but usually the definitive way this works even without special tooling support is by replaying the migrations to the state before the MR and then capture the diff for the target state at the "end" of the PR.

tomkarw (approved):
As a significant improvement in part of the project that as of now is still WIP, I think this MR is worth merging as-is?

@BlowaXD @mara-schulke

cdbrkfxrpt (approved):

PR #219, opened 2024-02-06
Improve nix devx + verify it in CI

super-secret-github-user:
The following changes were made:
-  add 2 new CI jobs which run under a new `Nix Buffers CI` workflow (we also cache the nix store locally)
-- `nix flake check` on ubuntu
-- `nix flake check` on mac
- enable all CI workflows to be manually triggered from the `actions` panel
- clean up `flake.nix`
- speed up nix builds by providing libgit and openssl, instead of building from source

`nix flake check` now does the following:
- ensure all `nix` files are correctly formatted
- verify we can build buffrs under nix


.envrc
```diff
@@ -0,0 +1 @@
+use flake
```
mara-schulke, line 1: Can you gitignore this?

super-secret-github-user: I'd rather not - if you're running helsing's nix-template (or direnv in general) that will automatically start the devshell with all of the project's dependencies + dev tools when you enter the directory.

flake.nix
```diff
@@ -5,27 +5,76 @@
...
-          src = ./.;
-          buildInputs = with pkgs; [ openssl ] ++ (pkgs.lib.lists.optionals (stdenv.isDarwin) [ darwin.apple_sdk.frameworks.SystemConfiguration ]);
+        nativeBuildInputs = with pkgs; [pkg-config];
+
+        darwin_frameworks = with pkgs.darwin.apple_sdk.frameworks; [
+          Security
+          SystemConfiguration
+        ];
+
+        dev_tools = with pkgs; [
```
j-baker, line 29: please can you camelCase and not snake case the variables?

flake.nix
```diff
@@ -5,27 +5,76 @@
...
         };
-        devShells.default = pkgs.mkShell {
-          nativeBuildInputs = with pkgs; nativeBuildInputs ++ [
-            cargo
-            rustc
-          ];
+      in rec {
+        # NB: if this does not build and you need to modify the file,
+        #     please ensure you also make the corresponding changes in the devshell
+        packages.default = naersk'.buildPackage ({
```
j-baker, line 48: any chance we could take this opportunity to build with Crane and not Naersk? I've seen a bunch of internal issues which I _think_ are triggered by building with Naersk and not Crane.

flake.nix
```diff
@@ -5,27 +5,76 @@
...
+        checks = {
+          builds = packages.default;
+          nix-files-are-formatted = pkgs.stdenvNoCC.mkDerivation {
+            name = "fmt-check";
+            dontBuild = true;
+            src = ./.;
+            doCheck = true;
+            nativeBuildInputs = with pkgs; [alejandra];
+            checkPhase = ''
+              alejandra -c .
```
j-baker, line 71: I think internally we're trying to use nixfmt and not alejandra. it'd be good to at least be consistent.

j-baker: also, https://github.com/NixOS/rfcs/pull/166

internal rationale for posterity:

```
The three main options are:
[nixpkgs-fmt](https://github.com/nix-community/nixpkgs-fmt): this is the most "official" of the bunch, maintained in the nix-community GitHub org and designed to format nixpkgs. It has a weird design choice though of "only expand, never collapse" which tends to result in long files, with short line lengths, during my testing.
[nixfmt](https://github.com/serokell/nixfmt): I tried this out after not liking the "only expand, never collapse" of nixpkgs-fmt. It was fine: fast, installed easily (with Nix, of course) and integrated with my editor and pre-commit easily. I've been using it on side projects a few weeks without issues.
[alejandra](https://github.com/kamadorueda/alejandra): less actively maintained, and resulted in the largest diffs when I ran it on a subset of nixpkgs, which suggests its preferred style is divergent from community norms.
```

super-secret-github-user: Happy to do so. It's worth mentioning that nixpkgs will use both, and alejandra is a direct response to nixfmt breaking some nixpkgs builds.

.envrc
mara-schulke: Yes, this is helsing nix-template specific which is why I asked everyone that had these in their PR until now (happened multiple times) to remove them as they don't serve a broader usage to the general public!

I am not convinced that local environment settings, such as your shell setup and QoL features, IDE settings or alike should live in an open source repository! 🙈 There are just too many different tools (for example for the env management: direnv, ondir, autoenv, shadowenv etc) for this and you clutter the repository over time if you don't have a strict policy on this.

I will put some time to write a `CONTRIBUTING.md` soon

mara-schulke (approved):
Except the comment RE gitignore, this lgtm!

flake.nix
super-secret-github-user: [Using this as a starting point](https://github.com/ipetkov/crane/blob/master/examples/quick-start/flake.nix)

flake.nix
```diff
@@ -5,27 +5,67 @@
...
+        devShells.default = pkgs.mkShell ({
+          buildInputs = nativeBuildInputs ++ devTools ++ dependencies;
+        } // envVars);
+
+        formatter = with pkgs;
+          writeShellApplication {
+            name = "nixfmt-nix-files";
+            runtimeInputs = [ fd nixfmt ];
+            text = "fd \\.nix\\$ | xargs nixfmt";
+          };
```
super-secret-github-user, line 50: Drawing attention to the new formatter script in case this is different to what will end up in helpkgs

flake.nix
piegamesde: If RFC 166 will be accepted then Nixpkgs will not use both. Alejandra raised the bar when it was created, but nowadays Nixfmt fulfills pretty much the same correctness guarantees. (Though be sure to use the `nixfmt-rfc-style` package or the RFC development branch.)

.nix/buffrs.nix
```diff
@@ -0,0 +1,41 @@
+{ pkgs, crane, rustToolchain, system, advisory-db, buildInputs
+, nativeBuildInputs, buildEnvVars }:
+let
+  craneLib = crane.lib.${system};
```
j-baker, line 4: i think you want to set the toolchain on this

.nix/buffrs.nix
```diff
@@ -0,0 +1,41 @@
+{ pkgs, crane, rustToolchain, system, advisory-db, buildInputs
```
j-baker, line 1: please can we not put nix stuff in a hidden folder?

.nix/libgit2.nix
```diff
@@ -0,0 +1,51 @@
+{ lib, stdenv, fetchFromGitHub, cmake, pkg-config, python3, zlib, libssh2
+, openssl, pcre, http-parser, libiconv, Security, staticBuild ?
+  stdenv.hostPlatform.isStatic
+  # for passthru.tests
+, libgit2-glib, python3Packages }:
+
+stdenv.mkDerivation rec {
+  pname = "libgit2";
+  version = "1.7.2";
```
j-baker, line 9: is there a reason that the libgit2 1.7.1 in nixpkgs is not sufficient?

.nix/libgit2.nix
```diff
@@ -0,0 +1,51 @@
...
+  passthru.tests = {
+    inherit libgit2-glib;
+    inherit (python3Packages) pygit2;
+  };
+
+  meta = with lib; {
+    description =
+      "Linkable library implementation of Git that you can use in your application";
+    homepage = "https://libgit2.org/";
+    license = licenses.gpl2Plus;
```
j-baker, line 47: gpl2Plus?

j-baker: ah, there's the linking exception. ok.

.nix/toolchain.nix
```diff
@@ -0,0 +1,7 @@
+{ rust-bin }:
+
+rust-bin.stable.latest.default.override {
+  extensions = [ "rust-src" "rust-analyzer" ];
+  targets = [ "arm-unknown-linux-gnueabihf" ];
```
j-baker, line 5: this seems fishy. i'd expect the default one to at least contain the appropriate target for the system

.nix/libgit2.nix
super-secret-github-user: pkg-config for openssl insists on 1.7.2

.nix/libgit2.nix
super-secret-github-user: I just copied the derivation from nixpkgs

.nix/buffrs.nix
super-secret-github-user: moved to a normal `nix` folder

.github/workflows/nix_ci.yml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: DeterminateSystems/nix-installer-action@main
+      - uses: DeterminateSystems/magic-nix-cache-action@main
+      - uses: DeterminateSystems/flake-checker-action@main
+      - name: Run `nix flake check`
+        run: nix flake check
+
+  check_flake_w_nix_on_mac:
+    runs-on: macos-latest
```
j-baker, line 22: MacOS runners cost $0.08/minute and your last build took 15 minutes. Let's make sure this job runs quickly if it is to be used when we merge, both for iteration reasons and cost reasons (if each buffrs mr takes 5 attempts, it feels unreasonable for this to cost $12.

.nix/libgit2.nix
j-baker: for openssl? might be worth instead downgrading the libgit2 dep in the lockfile. 1.7.2 was released exactly two days ago.

flake.nix
```diff
@@ -1,31 +1,95 @@
...
+        } // buildEnvVars);
+
+        formatter = with pkgs;
+          writeShellApplication {
+            name = "nixfmt-nix-files";
+            runtimeInputs = [ fd nixfmt ];
+            text = "fd \\.nix\\$ --hidden --type f | xargs nixfmt";
+          };
+
+        checks = ({
```
j-baker, line 77: i think you can just run `nixfmt -c .` for this whole thing?

super-secret-github-user: It will pick up non-nix files then

super-secret-github-user: The mac run now takes 9 mins if you need to rebuild the deps.

Building the deps took at least 5 mins. If you don't need to change them, you'll use the cache and ideally get an appropriate speedup.

.github/workflows/nix_ci.yml
super-secret-github-user: When we're pulling from cache, we can finish the job in 1m. So it looks like the mandatory cost of a job rn is 0.08$, and any extra is incurred when there is more work to do.

super-secret-github-user: I've now set the mac CI run to only happen after a merge

Issue #217, opened 2024-01-25
consider gix instead of libgit2


j-baker:
libgit2 is a fairly minor dependency for buffrs, but at a fairly high cost for users. git is in the default featureset, so:

1. Builds take a long time due to need to compile openssl, libgit2, etc (e.g. a cargo package using the buffrs crate will take 5 minutes to compile every time the right env vars change).
2. They can fail more easily on esoteric platforms. For example, I'm using cargo zigbuild to build python wheels, and this is complaining when trying to access zlib during the vendored openssl build. I can only assume it's gotten confused with its compilers in one place or another (maybe it's trying to use clang when it should be using gcc? not sure).

Instead of using libgit2 and all the extra cruft this entails, we could use gix instead, which I think supports the specific functionality we need. This would simplify the native dependency tree and reduce install times.

mara-schulke:
FWIW @j-baker you can disable the git feature!

vsiles:
I'm not familiar with these libs but that's a good way to get to know more of buffrs. I'll spend some time on it to assess if `gix` is a good substitute, and do the refactoring if that's the case. Thanks for flagging it

vsiles:
From what I understand of the usage of git2 in buffrs, it is mostly to check if the repo is "clean" (all files are either committed or ignored).

gix and gix_status don't seem to be "there yet":
- using these crates, I managed to get information on any tracked files with non committed changes
- however I can't find how to get information about untracked files. Looking at [the project status](https://github.com/Byron/gitoxide/blob/main/crate-status.md) it seems this is not supported at the moment.

The current code would complain if some untracked files are present. We could walk the file system and check if there are some files that are not tracked and not ignored to achieve our goal, but I don't know if gix can give me information about ignored files. I'll try to check that and be back at you.

tomkarw:
Completed in https://github.com/helsing-ai/buffrs/pull/220.

PR #220, opened 2024-02-08
Removing dependency to git2

vsiles:
This is related to https://github.com/helsing-ai/buffrs/issues/217

git2 is replaced by:
- in buffrs itself, we now call `git status --porcelain` directly, since `gix` is not able to give us information about untracked files.
- in the test-suite, we use `gix` instead of `git2`

The test suite has been updated to deal with some minor behaviour.

I run `time cargo build` before/after to compare since build time was an issue, here are the results on my laptop (M3):

- before:
```
    Finished dev [unoptimized + debuginfo] target(s) in 57.88s

________________________________________________________
Executed in   57.95 secs    fish           external
   usr time  185.65 secs    0.16 millis  185.65 secs
   sys time   55.63 secs    1.15 millis   55.63 secs
```

 - after:
 ```

    Finished dev [unoptimized + debuginfo] target(s) in 15.59s

________________________________________________________
Executed in   15.66 secs    fish           external
   usr time  102.08 secs    0.18 millis  102.08 secs
   sys time   12.89 secs    1.19 millis   12.89 secs
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ test = true
 default = ["build", "git", "validation"]
 build = ["dep:tonic-build", "dep:protoc"]
 validation = ["dep:anyhow", "dep:protobuf", "dep:protobuf-parse", "dep:diff-struct"]
-git = ["dep:git2"]
+git = []
 
 [dependencies]
 async-recursion = "1.0.5"
@@ -47,7 +47,6 @@ bytes = "1.0"
 clap = { version = "4.3", features = ["cargo", "derive"] }
 diff-struct = { version = "0.5.3", optional = true }
 flate2 = "1"
-git2 = { version = "0.18.1", features = ["vendored-openssl", "vendored-libgit2"], optional = true }
 hex = "0.4.3"
 home = "0.5.5"
 human-panic = "1"
@@ -75,6 +74,8 @@ assert_cmd = "2.0"
 assert_fs = "1.0"
 axum = { version = "0.7.2", default-features = false, features = ["tokio", "http1"] }
 fs_extra = "1.3"
+gix-tempfile = "13.1.0"
+gix = { version = "0.58.0", default-features = false}
 hex = "0.4.3"
 paste = "1.0.14"
 predicates = "3.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,7 +74,6 @@ assert_cmd = "2.0"
 assert_fs = "1.0"
 axum = { version = "0.7.2", default-features = false, features = ["tokio", "http1"] }
 fs_extra = "1.3"
-gix-tempfile = "13.1.0"
 gix = { version = "0.58.0", default-features = false}
 hex = "0.4.3"
 paste = "1.0.14"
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: Cargo.toml [hunk 1]
```diff
@@ -75,6 +74,8 @@ assert_cmd = "2.0"
 assert_fs = "1.0"
 axum = { version = "0.7.2", default-features = false, features = ["tokio", "http1"] }
 fs_extra = "1.3"
+gix-tempfile = "13.1.0"
```
vsiles, line 77: For some reason without this, the version used was `13.0.0` and test failed to build with a missing function definition.

src/command.rs
```diff
@@ -195,6 +197,51 @@ pub async fn package(directory: impl AsRef<Path>, dry_run: bool) -> miette::Resu
         ))
 }
 
+#[cfg(feature = "git")]
```
vsiles, line 200: In this function, I tried to emulate the old behavior that would not `bail` if the command failed to run altogether.

mara-schulke:
Hi, is it possible to use gix for this? If not we can remove the feature altogether. Parsing cli outputs is super error prone and I would like to avoid this.

vsiles:
As I mention in the issue discussion, `gix` does not yet allow us to do what `git2` was doing: `gix` does not currently have information about untracked files, so we could only complain if some tracked files have changes.
That being said, I took care of calling `git status --porcelain` because it is made exactly for that:

> --porcelain[=<version>]
> Give the output in an easy-to-parse format for scripts. This is similar to the short output, but will remain stable across Git > versions and regardless of user configuration. See below for details.

That leaves as multiple options:

1.  we remove this check altogether
2. we use `gix` and accept to ignore untracked file (for now at least)
3. we use `git` but don't parse the output. If the output is empty, the repo is clean, otherwise it is not and we print the output as is
4. we trust the `--porcelain` option and parse the output

I don't have a strong opinion here, so I'll implement what you think is best

src/command.rs
```diff
@@ -30,6 +30,8 @@ use async_recursion::async_recursion;
 use miette::{bail, ensure, miette, Context as _, IntoDiagnostic};
 use semver::{Version, VersionReq};
 use std::{env, path::Path, str::FromStr};
+#[cfg(feature = "git")]
+use tokio::process;
```
tomkarw, line 34: I'd just qualify the import in line 202 and limit the amount of `#[cfg(feature = "git")]` spread across the codebase.

tomkarw:
I agree that using `git status --porcelain` will be robust enough, so I vote for option 4.

That said, I'm fuzzy on why we care about uncommitted changes in the first place. Maybe it's a chance to skim some overreaching features fat.

Cargo.toml
tomkarw: What are the steps to reproduce this issue? I commented this line out, run `cargo test`/`cargo test --all-features` and it seems to work.

vsiles:
Yes, I think I'm too new to have enough context here ;) I'm leaving that decision to @mara-schulke and you

mara-schulke:
Okay, so im okay with option 4 *but*: I would not error when the parsing fails. So I would at most emit a warning here that git could not be found / the output is malformed. Currently your implementation raises errors which would potentially abort the call.

If you make it best-effort (say: skip this and emit a warning line to the user) im okay with using `--porcelain`! 😊

vsiles:
Should be good for review now. Thanks for all the input.

tomkarw (approved):
LGTM, let's merge

Issue #102, opened 2023-10-06
Artifactory registry does not enforce origin/domain


asmello:
Currently the implementation of the Artifactory registry `download()` function does not use the base URL that was provided when it was initialised, and instead just makes a request to whatever URL the dependency provides with it:

https://github.com/helsing-ai/buffrs/blob/7467dbc7b6ef76b743e42b1a701f11357729d22c/src/registry/artifactory.rs#L101-L115

This is a bug. Either the registry implementation should be agnostic to the specific URL the dependency is being downloaded from, OR it should enforce that it is only used to download packages from the registry URL it was configured with. These are mutually exclusive design choices.

PR #221, opened 2024-02-11
Big change

khanism:
Enforcing configured registry URL (see https://github.com/helsing-ai/buffrs/issues/102)

### DIFF
--- a/src/registry/artifactory.rs
+++ b/src/registry/artifactory.rs
@@ -72,7 +72,7 @@ impl Artifactory {
 
             let path = dependency.manifest.registry.path().to_owned();
 
-            let mut url = dependency.manifest.registry.clone();
+            let mut url = self.registry.clone();
             url.set_path(&format!(
                 "{}/{}/{}/{}-{}.tgz",
                 path,


mara-schulke:
Hi, thank you for you contribution!

Did you encounter an issue where this is causing trouble? 

I think in the current state of the manifest (e.g. the manifest contains an absolute uri on where to retrieve the buffrs package from) it is dangerous to set the uri to point to self.registry

I think as @asmello pointed out it would be valuable to not implement the registry as a struct containing a url but rather always use the URI from the manifest.

---

My reasoning behind this is:

Imagine two dependencies in two distinct registries which are both proprietary (`a` in registry `company-a` and `b` in registry `company-b`).

`a` depends on `b` but

mara-schulke:
Hi, thank you for you contribution!

Did you encounter a situation where this is causing trouble? 

I think in the current state of the manifest (e.g. the manifest contains an absolute uri on where to retrieve the buffrs package from) it is dangerous to set the uri to point to self.registry

I think as @asmello pointed out it would be valuable to not implement the registry as a struct containing a url but rather always use the URI from the manifest.

###### My reasoning behind this:

Imagine two dependencies in two distinct registries which are both proprietary:

- api `a` in registry `company-a`
- lib `b` in registry `company-b`
- `a` depends on `b@1.0.0`

If someone wants to install api `a` now in his project he would declare a dependency on it:

```toml
[dependencies]
a = { registry = "https://company-a.jfrog.io/artifactory", version = "1.0.0", repository = "test" }
```

When `buffrs install` is being run this would look the following (written in pseudo code):

```
a = download("https://company-a.jfrog.io/artifactory/test/a-1.0.0.tgz")
manifest_a = unpack(a)

for dependency in manifest_a:
    dep = download("{dependency.registry}/{dependency.repository}-{dependency.version}.tgz")
    dep_manifest = unpack(dep)
    // infinetly traverse the tree
```

This algorithm makes a call to `https://company-b.jfrog.io/artifactory/test/b-1.0.0.tgz` to resolve `b` which is correct.

Your change (as far as I understand) would change the above algorithm to the following:

```
a = download("https://company-a.jfrog.io/artifactory/test/a-1.0.0.tgz")
manifest_a = unpack(a)

for dependency in manifest_a:
    dep = download("https://company-a.jfrog.io/artifactory/{dependency.repository}-{dependency.version}.tgz")
    dep_manifest = unpack(dep)
    // infinetly traverse the tree
```

Which would try to resolve the library `b` in the registry of `company-a`. This seems incorrect to me.

mara-schulke:
Thank you a lot @khanism! If you need anything, let me know 😊

PR #222, opened 2024-02-13
Eg/add sheilds to indicate build status

super-secret-github-user:
Added shields to indicate the status of nix CI.

Mac builds only run after a merge. This lets us see at a glance whether something has gone wrong.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -11,3 +11,6 @@ license-files = [{ path = "LICENSE", hash = 0xbd0eed23 }]
 [bans]
 multiple-versions = "warn"
 wildcards = "deny"
+
+[advisories]
+ignore = ["RUSTSEC-2024-0013"]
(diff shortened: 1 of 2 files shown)


super-secret-github-user:
After speaking with @j-baker, it seems that it is fine to proceed in the presence of the cargo deny issue as buffrs does not use the bits of libgit2 with the vulnerability

super-secret-github-user:
NB: I have set myself a slack reminder to see if 1.7.2 is in nixpkgs for one week from now. So, I will remove the ignore when nixpkgs ship the latest version of libgit2

deny.toml
```diff
@@ -11,3 +11,6 @@ license-files = [{ path = "LICENSE", hash = 0xbd0eed23 }]
 [bans]
 multiple-versions = "warn"
 wildcards = "deny"
+
+[advisories]
+ignore = ["RUSTSEC-2024-0013"]
```
Tpt, line 16: Why not just upgrade libgit2?

Tpt (approved):

deny.toml
Tpt: Sorry, missed your comment on the main thread. Closing.

vsiles: FYI we're dropping libgit2 soon (see https://github.com/helsing-ai/buffrs/pull/220)

Issue #143, opened 2023-10-20
Buffrs library pulls in a lot of dependencies that are not needed.


xfbs:

mara-schulke:
@xfbs we already make use of cargo features to only inline the bare minimum of features for each target. Maybe we can improve them, but i dont think we need to split crates here?

j-baker:
I'm with @xfbs on this. I'm trying to use buffrs, and I'm finding it's pulling in (in my cargo build):

```
libz-sys
libssh-sys
libgit2-sys
```

Each of these adds a few seconds to my compile, are duplicated due to buffrs being a build and normal dependency, and worse, the requirements for rebuilding are highly frequent on environment changes. Essentially, running 'cargo clippy' from a vscode shell and then running it from iterm is enough to cause these deps to require a full rebuild (20 seconds overall) whereas without them, the build would only require compiling my crates.

My understanding is that it's not easy to have a dependency be default included with a binary target, but optional if depended on as a lib.

At present, I think I can work around this by depending on buffrs as:

```
buffrs = { version = "0.7.2", default_features = false, features = ["build"]}
```

or similar. But it feels unfortunate that using buffrs as documented needlessly slows down your development, and my experience is that crate boundaries are the correct way to ensure appropriate dependency choice. In particular, I think the thing that is being traded off here is:

- for a cli, you basically can use any dependency you want, because it's compile once, use many.
- for a library, you want to use the absolute minimum set of dependencies, because it can add requirements for usage which aren't desirable. For some examples, libgit2 and libssh add dependencies on a native libz being present, clap adds a very recent MSRV, etc.

xfbs:
@j-baker, I think I had identified the similar issues as you. Even thought we already do some feature-gating as @mara-schulke suggested, we might have a fundamental issue right now.

## Problem definition

Under the same `buffrs` namespace, we want to house two relatively different things:

- the `buffrs` library, which should be as bare-bones as possible (only compiles protobuf; very minimal, carefully-chosen dependencies because this is rebuilt frequently),
- the `buffrs` CLI, which should be nice (using `clap`; having a lot of features out-of-the-box, err on the side of having more features and dependencies because they only need to be built once).

In other words: the default feature (and thus dependency) sets of these two use-cases are different. 

Ideally, you want to be able to add this to your project's dependencies and be sure to get `buffrs` with the minimal feature set you need to use it productively (aka, it has everything it needs to compile protocol buffer files, and nothing more):

```toml
[...]

PR #240, opened 2024-04-25
chore: transform the repo into a workspace to prepare splitting into crates

poliorcetics:
Related to https://github.com/helsing-ai/buffrs/issues/143


mara-schulke:
Hi @poliorcetics, I believe this effort is not the most sensible as we are not working on the registry anyway. It would be much easier to just remove the registry implementation all together instead of adapting the project setup to support it. 

What's your thoughts?

poliorcetics:
It's not for the registry, I want to extract `buffrs-cli` and `buffrs-core` and maybe others. I will make an MR to remove the registry though, that will greatly simplify things

PR #244, opened 2024-05-09
Add rust tests to nix checks

DavidWHelsing:
Rust tests weren't current running as part of nix checks. This PR adds `nextest` to our nix checks run.

I ignored two of the integration tests which require CA root certificates to be setup because they'll need extra changes and I wanted to keep the PRs small. Another PR will follow

### DIFF
--- a/nix/buffrs.nix
+++ b/nix/buffrs.nix
@@ -44,7 +44,7 @@ in {
       partitions = 1;
       partitionType = "count";
       # Ignore tutorial and publish tests because they need TLS certificates 
-      # which doens't yet work in our next setup
+      # which doens't yet work in our nix setup
       cargoNextestExtraArgs =
         "--filter-expr 'all() - test(=cmd::tuto::fixture) - test(=cmd::publish::fixture)'";
     });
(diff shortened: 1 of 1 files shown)


DavidWHelsing:
The `Clippy` failure seems unrelated to my changes in this PR.
It looks like a positive catch that was previously undetected but is now detected with a new version of `clippy`

nix/buffrs.nix
```diff
@@ -37,5 +37,16 @@ in {
...
     # Audit licenses
     buffrs-deny = craneLib.cargoDeny { inherit src; };
+
+    # Rust unit and integration tests
+    buffers-nextest = craneLib.cargoNextest (commonArgs // {
+      inherit cargoArtifacts;
+      partitions = 1;
+      partitionType = "count";
+      # Ignore tutorial and publish tests because they need TLS certificates 
+      # which doesn't yet work in our nix setup
```
mara-schulke, line 47: Lets not ignore them, you can likely add the following to the `cargoNexttest` args:

```
SSL_CERT_FILE = "${pkgs.cacert}/etc/ssl/certs/ca-bundle.crt";
```

mara-schulke: This will prob make there cert available to the tests

DavidWHelsing: Ooooh! I didn't realize you can just export env vars in there. That's so much easier!

mara-schulke (approved):

PR #245, opened 2024-05-09
Fix Clippy warnings

DavidWHelsing:
Fixing some clippy warnings that are causing CI failures.

None of these seem to appear in public API so they shouldn't need a version bump

### DIFF
--- a/src/registry/cache.rs
+++ b/src/registry/cache.rs
@@ -16,15 +16,9 @@ use std::path::PathBuf;
 
 use bytes::Bytes;
 use miette::{miette, Context, IntoDiagnostic};
-use semver::VersionReq;
-use thiserror::Error;
 use tokio::fs;
 
 use crate::{manifest::Dependency, package::Package};
-
-#[derive(Error, Debug)]
-#[error("{0} is not a supported version requirement")]
-struct UnsupportedVersionRequirement(VersionReq);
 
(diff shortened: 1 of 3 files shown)


src/registry/cache.rs
```diff
@@ -16,16 +16,10 @@ use std::path::PathBuf;
...
 use miette::{miette, Context, IntoDiagnostic};
-use semver::VersionReq;
-use thiserror::Error;
 use tokio::fs;
 
 use crate::{manifest::Dependency, package::Package};
 
-#[derive(Error, Debug)]
-#[error("{0} is not a supported version requirement")]
-struct UnsupportedVersionRequirement(VersionReq);
```
DavidWHelsing, line 27: This type is also defined in `src/registry/mod.rs` so this is just a duplicate declaration

tests/cmd/tuto/in/main.rs
```diff
@@ -2,6 +2,7 @@ mod protos {
     tonic::include_proto!("buffrs");
 }
 
+#[allow(dead_code)]
```
DavidWHelsing, line 5: This code is only defined in the code that's compiled as part of the tutorial test. The warning only pops up as part of `cargo build` in the test output

mara-schulke (approved):

PR #246, opened 2024-05-09
Simplify project root detection in tutorial test

DavidWHelsing:
The project source code detection in the tutorial test was using `gix` to detect root of the rust project by looking for the `.git` directory. 
As far as I can tell this is unnecessary since rust tests are always executed from the project root. So the working path should already be the correct directory.
Even if the tests were somehow manually executed from a different path I don't think the existing code would have worked since it was joining current working directory with `crate::parent_directory!()` which would be the relative path of the directory in which our code file is located. That would be incorrect for any other working directory.

This assumption might not work in a cargo workspace but I think that this code would already not work in a workspace and this project isn't using a workspace right now.

Simplifying this code also allows us to remove the dependency on on `gix`. Since this is a dev dependancy change there should be no need for a version bump


tests/cmd/tuto/mod.rs
```diff
@@ -132,21 +132,19 @@ fn fixture() {
...
-            let git_root = git_repo
-                .path()
-                .parent()
-                .expect("failed to move from path to .git to its parent")
-                .to_str()
-                .expect("failed to convert path git root to string");
-            dbg!(git_root);
+            let project_root = std::env::current_dir().expect("Failed to get working directory");
+            let project_root_str = project_root.to_str().expect("Failed to convert to str");
+            dbg!(project_root_str);
```
mara-schulke, line 137: can we remove this?

tests/cmd/tuto/mod.rs
```diff
@@ -132,21 +132,19 @@ fn fixture() {
...
-                .join(crate::parent_directory!());
-            let git_repo = gix::discover(path).expect("unable to find git root");
-            let git_root = git_repo
-                .path()
-                .parent()
-                .expect("failed to move from path to .git to its parent")
-                .to_str()
-                .expect("failed to convert path git root to string");
-            dbg!(git_root);
+            let project_root = std::env::current_dir().expect("Failed to get working directory");
```
mara-schulke, line 135: this is likely going to break

e.g. the gix code was added for a reason by @qsantos iirc

mara-schulke: but: if the tests pass, im happy to merge

mara-schulke:
Seems like the tests pass! Just beware that you now have a clippy error in ci

Also I think this code used to be in there to make `cargo test` executable in subdirectories / crates

tests/cmd/tuto/mod.rs
qsantos: There was no detection for the actual root of the project. Maybe this was added since, so manually navigating to the parent directory is not needed anymore?

DavidWHelsing: This is using current working directory. So that's the directory from which it's executed.
Invoking this test using `cargo test` or `cargo nextest run` will execute it from the root of the project even if you are in a subdirectory.

If there is a reason this was here I am happy to keep it. I was trying to remove it because it wasn't working under nix. Alternatively we could just depend on git inside of the nix test

mara-schulke:
@DavidWHelsing please go ahead!! Im happy to accept whatever changes improve code quality while keeping the test passing 😊

PR #253, opened 2024-06-26
Add .buffrs/config.toml

dgehriger:
Based on upstream version 0.8.1 but with following additions:

- Introduction of a configuration file in _.buffrs/config.toml_ (in the current working directory, or any parent)
- Config file allows specifying:
  - List of registry URLs and their names
  - A default registry

Example:

```toml
[registries]
some_org = "https://artifactory.example.com/artifactory/some-org"

[registry]
default = "some_org"
```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,0 +1,198 @@
+// Copyright 2023 Helsing GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::registry::RegistryUri;
+use miette::{bail, ensure, miette, Context, IntoDiagnostic};
+use std::{
+    collections::HashMap,
+    path::{Path, PathBuf},
+    str::FromStr,
+};
+
+/// Representation of the .buffrs/config.toml configuration file
+///
+/// # Example
+///
+/// ```toml
+/// [registries]
+/// some_org = "https://artifactory.example.com/artifactory/some-org"
+///
+/// [registry]
+/// default = "some_org"
+/// ```
+///
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Config {
+    /// Path to the configuration file
+    config_path: Option<PathBuf>,
+
+    /// Default registry to use if none is specified
+    default_registry: Option<String>,
+
+    /// List of registries
+    registries: HashMap<String, RegistryUri>,
+}
+
+impl Config {
+    /// Create a new configuration with default values
+    /// # Arguments
+    /// * `cwd` - Starting directory to search for the configuration file
+    ///
+    pub fn new(cwd: Option<&Path>) -> miette::Result<Self> {
+        match Self::locate_config(cwd) {
+            Some(config_path) => Self::new_from_config_file(&config_path),
+            None => Ok(Self {
+                config_path: None,
+                default_registry: None,
+                registries: HashMap::new(),
+            }),
+        }
+    }
+
+    /// Resolve the registry URI from the configuration
+    ///
+    /// # Arguments
+    /// * `registry` - The registry name or URI to resolve
+    ///
+    /// # Returns
+    /// The resolved registry URI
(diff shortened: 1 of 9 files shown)


poliorcetics:
I prefer `.config/buffrs/config.toml` to avoid the proliferation of dotfiles in a repo (for example in the Rust community, cargo-nextest already does this)

tomkarw:
+1 one on what @poliorcetics said.

> I don't have a strong opinion, and am happy to change the PR accordingly. However, I don't actually see any tool other than nextest using a _.config/*_ folder for its settings. Can you please point out a few to me?

Did you try running it on your local machine? I get:
```sh
➜  ~ ls ~/.config | wc -l
66
```
Some notable entries are `git`, `Code`, `helix`, `libreoffice`, `mozzila`, `tmux`.

Also, I love XKCD as much as the next guy, but I don't think it applies here. We're not the ones creating a new standard, We're just trying to coalesce around an existing one.

dgehriger:
> Also, I love XKCD as much as the next guy, but I don't think it applies here. We're not the ones creating a new standard, We're just trying to coalesce around an existing one.

@tomkarw : your requested change has been implemented. Would you please re-review or permanently reject this PR to clarify its status?

src/config.rs
```diff
@@ -0,0 +1,200 @@
...
+    /// * `cwd` - Starting directory to search for the configuration file
+    ///
+    /// # Returns
+    /// Some(PathBuf) if the configuration file is found, None otherwise
+    fn locate_config(cwd: Option<&Path>) -> Option<PathBuf> {
+        if let Some(cwd) = cwd {
+            let mut current_dir = cwd.to_owned();
+
+            loop {
+                let config_path = current_dir.join(".config/buffrs/config.toml");
```
tomkarw, line 123: Slight misunderstanding, the config should be placed in a user home directory, i.e. `~/.config/buffrs`. The exact path can be customized, so there's a helper crate to resolve the actual path:
```suggestion
                use etcetera::BaseStrategy;

                let config_path = etcetera::choose_base_strategy()?.config_dir().join("buffrs/config.toml");
```

and in `Cargo.toml`
```toml
etcetera = "0.8.0"
```

Then, you can actually remove the looping logic and the whole `locate_config` function and just use the above snippet in `Config::new()`.

src/config.rs
```diff
@@ -0,0 +1,200 @@
...
+    pub fn new(cwd: Option<&Path>) -> miette::Result<Self> {
+        match Self::locate_config(cwd) {
+            Some(config_path) => Self::new_from_config_file(&config_path),
+            None => Ok(Self {
+                config_path: None,
+                default_registry: None,
+                registries: HashMap::new(),
+            }),
+        }
+    }
```
tomkarw, line 61:
```suggestion
    pub fn from_config() -> miette::Result<Self> {
        let config_path = etcetera::choose_base_strategy()?.config_dir().join("buffrs/config.toml");
        if config_path.exists() {
            Some(config_path) => Self::new_from_config_file(&config_path),
            None => Ok(Self {
                config_path: None,
                default_registry: None,
                registries: HashMap::new(),
            }),
        }
    }
```

Could probably be simplified further, not sure we need both `new_from_config_file` and `from_config`.

tomkarw (changes requested):
Almost there, let's just adjust where the config is stored.

src/config.rs
dgehriger: @tomkarw : placing the config in the user's home directory actually defies the very purpose of this PR. In our org, we need to pre-configure the registry URIs on a per-repo basis. So this definitely needs to live with the repo, not the user.

Without this feature, I prefer to close the PR.

Let me know your opinion -- if we agree then I'll address your other concerns below.

mara-schulke: Hi @dgehriger i am sorry that I was unresponsive with regards to your PRs and I'm happy to jump on a call if you want to reopen them to figure out how we can merge your work.

For this specific change I do agree that the configuration file would need to be included in the repository itself rather than the home directory.

Also, naively asked, how does this affect dependency resolution?

dgehriger: @mara-schulke : no problem, we all have our projects and timelines. I'd actually love to get on a call with you (maybe next week ?) to sync up. We are using your wonderful tool inside our organization, but it's true that I had to fork it to make sure it matches our workflows. It would be wonderful if, instead of us maintaining a diverging version of `buffrs`, we could bring this together. But it's not a hard requirement on my side.

Can you drop me a mail on dgehriger@globusmedical.com and suggest a time that suits you for a video call during (ideally sometimes between 9 - 12 am Central European Time -- I'm located close to you, in Lausanne, Switzerland)?

dgehriger: @tomkarw : since your suggestion was regarding _~/.config/..._ and doesn't apply to repository configuration files, I've reverted the change in the latest commit.

poliorcetics:
I'll reiterate, I much prefer `<repo>/.config/buffrs.toml`, spamming repos with dotfiles is not my idea of quality

Issue #185, opened 2023-12-06
Have `buffrs init dir` use dir instead of the current working directory


qsantos:
I was a bit surprised when `buffrs init --lib units` created `Proto.ml` and `proto/` in the current working directory instead of creating a `units/` directory, like Cargo would do.

The [second blog post](https://blog.helsing.ai/buffrs-a-package-manager-for-protocol-buffers-2-2-9f001faef6dc) actually uses `buffrs init --lib units` and suggests it works (or worked) like Cargo.

The current behavior is a bit annoying, since it mean you need to `mkdir` and `cd` into the directory before running `init`. If that's okay, I would like for `buffrs init` to work as closely as `cargo init` as possible, even initializing a git repository if there is none. It might not be very useful for implementations, but libraries could definitely use that.

vsiles:
Just to clarify the possible behaviors as we might want to distinguish package names from directory names. Right now:
- `buffrs init` in a directory named `foo` will initialize a `Proto.toml` with package name set to `foo` in the current directory `foo`
- `buffrs init bar` in a directory named `foo` will initialize a `Proto.toml` with package name set to `bar` in the current directory `foo`

First option is to go the `new / lib` way:
- `buffrs new` requires a mandatory package name and will create a buffrs project in a directory of that name, with a package name of the same name. I'd require the directory not to already exist to avoid having to check its content.
- `buffrs init` without a package name will perform like it does now, using the name of the current directory and setting things up in `.`
- `buffrs init package_name` will behave just like `buffrs new package_name`. We might allow the directory to already exist in that case. Unsure.

Now I read that the `package` section of `Proto.toml` files might go away. Would we keep the `name` around in that case, or use the directory name instead ?

PR #256, opened 2024-07-04
Add `buffrs new` command to init a package in a non-existing folder

gbouv:
https://github.com/helsing-ai/buffrs/issues/185


src/command.rs
```diff
@@ -139,7 +177,7 @@ pub async fn add(registry: RegistryUri, dependency: &str) -> miette::Result<()>
         .push(Dependency::new(registry, repository, package, version));
 
     manifest
-        .write()
+        .write(None)
```
mara-schulke, line 180: Can we do one method `write` and another one called `write_at` (or something along the lines) for the one taking a path instead of using a optional parameter?

src/command.rs
```diff
@@ -77,6 +81,40 @@ pub async fn init(kind: Option<PackageType>, name: Option<PackageName>) -> miett
     Ok(())
 }
 
+/// Initializes a project with the given name in the current directory
+pub async fn new(kind: Option<PackageType>, name: PackageName) -> miette::Result<()> {
+    let mut package_dir: PathBuf = ".".into();
+    package_dir.push::<String>(name.clone().into());
```
mara-schulke, line 87: What is the rationale against just constructing a pathbuf directly from the package name?

mara-schulke (commented):
Overall looks good to me! Could you add the synopsis/docs to both the README and the book before we merge?

src/command.rs
gbouv: Hmm, yeah nothing really, I'll change it

mara-schulke (approved):

PR #257, opened 2024-07-05
Allow dependencies in `Proto.lock` to be updated when manifest changes

fredrb:
This PR allows `buffrs install` to make changes to the lockfile when proto dependency versions were upgraded in the manifest. 

This code change is sufficient now because `buffrs` only support pinned versions in the manifest, so we can rely on the manifest always having the desired version. However, whenever we start supporting semantic version upgrades this logic might need to change. 

## How I validated that my change works

- [x] check that an upgrade worked by bumping `my-protos@0.0.1` to `my-protos@0.0.2` succeeded on `buffrs install` 
- [x] check that an sub-dependencies were upgraded by bumping `my-protos-with-deps@0.0.1` to `my-protos-with-deps@0.0.2` succeeded on `buffrs install` 
- [ ] automated testing

## Open Item

I wasn't able to load contents from cache in case we are ignoring the locked dependency and downloading a new version because `Cache` receives a `FileRequirement` which can be created from `LockedPackage` but can't be created from `Package`. I assume this could be fixed with `impl From<Package> for FileRequirement { ... }`.


src/resolver.rs
```diff
@@ -179,7 +175,9 @@ impl DependencyGraph {
                 })?;
 
             let package = registry
-                .download(dependency.with_version(&local_locked.version))
+                // fetch the version using manifest directly. This works now
+                // because buffrs only supports pinned versions
```
mara-schulke, line 179: I mean technically this is okay, since the entire version resolution is subject to change soon

mara-schulke: Nonetheless its a bit dangerous 😁

mara-schulke:
It would be great if we could write a snapshot test for this but im happy to get this merged! Its a big QoL improvement 😊

src/resolver.rs
fredrb: Do you want me to add a TODO here to remember this whenever dynamic versioning is implemented?

mara-schulke (approved):

mara-schulke:
Thank you!

PR #262, opened 2024-07-15
Local Dependencies

mara-schulke:
This introduces the ability to do the following in `Proto.toml` files:

```
[dependencies]
some-api = { path = "./path/to/some-api" }
```

## Notes

- Publishing any project with local dependencies is *not* possible
- This should be used primarily / only in consuming projects -> e.g. server impls
- This MR bumps the edition to `0.9.0`


src/resolver.rs
```diff
@@ -1,31 +1,64 @@
...
+        depends_on: Vec<PackageName>,
+    },
+    /// A resolved depnedency that is located on the filesystem
+    Local {
+        /// The materialized package that was created from the buffrs package at the given path
+        package: Package,
+        /// Location of the requested package
+        path: PathBuf,
+        /// Packages that requested this dependency (and what versions they accept)
+        dependants: Vec<Dependant>,
```
Tpt, line 42: What about just a `Vec<String>`? I believe the version range is always going to be `*`

src/resolver.rs
```diff
@@ -1,31 +1,64 @@
...
+
+impl ResolvedDependency {
+    pub(crate) fn package(&self) -> &Package {
+        match self {
+            Self::Remote { package, .. } => package,
+            Self::Local { package, .. } => package,
+        }
+    }
+
+    pub(crate) fn depends_on(&self) -> &Vec<PackageName> {
```
Tpt, line 56: nit: `&[PackageName]` (avoids an extra indirection)

Tpt:
```suggestion
    pub(crate) fn depends_on(&self) -> &[PackageName] {
```
(avoids an extra indirection)

src/registry/cache.rs
```diff
@@ -35,11 +38,18 @@ impl LocalRegistry {
 
     /// "Downloads" a package from the local filesystem
     pub async fn download(&self, dependency: Dependency) -> miette::Result<Package> {
+        let DependencyManifest::Remote(ref manifest) = dependency.manifest else {
+            return Err(miette!(
+                "unable to serialize version of local dependency ({})",
+                dependency.package
+            ));
+        };
```
Tpt, line 46: Nit: not useful code, `dependency_version_string` already does the check

Tpt (approved):

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "buffrs"
-version = "0.8.1"
+version = "0.9.0"
```
poliorcetics, line 3: IMO the version bump should be done in another mr to separate concerns

src/registry/artifactory.rs
```diff
@@ -197,6 +200,20 @@ impl Artifactory {
...
+            .dependencies
+            .iter()
+            .find(|d| d.manifest.is_local())
+        {
+            return Err(miette!(
+                "unable to public {} to artifactory due to local dependency {}",
+                package.name(),
+                local.package
+            ));
+        }
```
poliorcetics, line 215: we could also print the total number of local deps to avoid people forgetting one, getting the error again and restarting the cycle

src/resolver.rs
mara-schulke: Although this is correct im in favor of using types to communicate what is inside of the `String`

src/registry/cache.rs
mara-schulke: Yes but that is not technically part of the contract that `dependency_version_string` communicates – ie. although this is the current behavior the check for local dependencies is implicitly happening, rather than explicitly

Cargo.toml
mara-schulke: I kind of agree, although this change _is directly tied_ to a manifest change specifically and it should not be rolled out without the version bump

src/registry/artifactory.rs
mara-schulke: Good hint! Having to rerun buffrs publish over and over just to get a new error every time is a bad ux!

Issue #176, opened 2023-12-04
Improve and stabilize the protobuf validation


mara-schulke:
cc @xfbs 

We need to roll out more linting rules and restructure the validation implementation eventually.

@xfbs could you guide tom on this?

PR #264, opened 2024-08-07
issue-176: add lint to keep service defs out of lib type packages

Leopard2A5:
Addressing one of the todos in #176: add a lint for lib type packages emitting a warning when the packages declares any services.

Screenshot 2024-08-07 at 18 55 21

### DIFF
--- a/src/validation/rules/lib_package.rs
+++ b/src/validation/rules/lib_package.rs
@@ -1,0 +1,80 @@
+use super::*;
+
+#[derive(Debug, Clone, Copy)]
+pub struct LibPackage;
+
+impl Rule for LibPackage {
+    fn rule_info(&self) -> &'static str {
+        "Make sure that lib packages don't contain service definitions."
+    }
+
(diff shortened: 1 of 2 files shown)


src/validation/rules/lib_package.rs
```diff
@@ -0,0 +1,80 @@
...
+#[derive(Debug, Clone, Copy)]
+pub struct LibPackage;
+
+impl Rule for LibPackage {
+    fn rule_info(&self) -> &'static str {
+        "Make sure that lib packages don't contain service definitions."
+    }
+
+    fn rule_level(&self) -> Level {
+        Level::Warning
```
mara-schulke, line 12: Would suggest erroring?

mara-schulke (approved):

src/validation/rules/lib_package.rs
Leopard2A5: done :)

PR #265, opened 2024-08-08
Rene/issue 176 lints (package names and folder structure)

Leopard2A5:
Implement lint that makes sure the package hierarchy is mirrored by the folder structure.

Screenshot 2024-08-08 at 14 56 12

### DIFF
--- a/src/validation/rules/lib_package.rs
+++ b/src/validation/rules/lib_package.rs
@@ -8,10 +8,6 @@ impl Rule for LibPackage {
         "Make sure that lib packages don't contain service definitions."
     }
 
-    fn rule_level(&self) -> Level {
-        Level::Error
-    }
-
     fn check_package(&mut self, package: &Package) -> Violations {
         let package_name = &package.name;
         package
--- a/src/validation/rules/package_hierarchy.rs
+++ b/src/validation/rules/package_hierarchy.rs
@@ -144,16 +144,16 @@ mod tests {
             PackageHierarchy.check_package(&package),
             &[
                 Violation {
-                rule: "PackageHierarchy".into(),
-                level: Level::Error,
-                message: violation::Message {
-                    message:
-                        "expected file proto/not_subpackage/file.proto to live in proto/subpackage."
-                            .into(),
-                    help: EXPECTED_HELP.into(),
-                },
-                location: Default::default(),
-                info: PackageHierarchy.rule_info().into(),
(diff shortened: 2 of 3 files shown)


src/validation/rules/package_hierarchy.rs
```diff
@@ -0,0 +1,172 @@
...
+mod tests {
+    use super::*;
+
+    static EXPECTED_HELP: &str = "Package names should be mirrored in folder structure, eg mypackage.subpackage should live in proto/subpackage.";
+
+    #[test]
+    fn file_name_should_not_matter() {
+        let package = Package {
+            name: "root".into(),
+            files: vec!["proto/file_not_called_root.proto".into()],
```
mara-schulke, line 78: This is actually fine!

mara-schulke: Ie. only subfolders in the proto folder denote subpackages

proto/*.proto -> pkg x;

proto/y/*.proto -> pkg x.y;

src/validation/rules/package_hierarchy.rs
```diff
@@ -0,0 +1,172 @@
...
+                "proto/not_subpackage/file.proto".into(),
+                "proto/foo/bar/file.proto".into(),
+            ],
+            entities: Default::default(),
+        };
+        assert_eq!(
+            PackageHierarchy.check_package(&package),
+            &[
+                Violation {
+                rule: "PackageHierarchy".into(),
```
mara-schulke, line 147: formatting?

src/validation/rules/lib_package.rs
```diff
@@ -8,10 +8,6 @@ impl Rule for LibPackage {
         "Make sure that lib packages don't contain service definitions."
     }
 
-    fn rule_level(&self) -> Level {
```
mara-schulke, line 11: What's the implication / reason behind this change?

mara-schulke (approved):

src/validation/rules/package_hierarchy.rs
Leopard2A5: done :)

src/validation/rules/lib_package.rs
Leopard2A5: That the default impl of `Rule` already returns `Level::Error`, so going with the default and not writing code that has no effect beyond the default.

PR #268, opened 2024-09-24
Update crane version to latest

ulucs:
Hello! In newer `crane` versions, `crane.lib.${system}` has been removed; so I'm creating this PR to:
- Replace the `crane.lib.${system}` usage with the equivalent `(crane.mkLib pkgs)`
- Update the `crane` version

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -2,10 +2,7 @@
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
     flake-utils.url = "github:numtide/flake-utils";
-    crane = {
-      url = "github:ipetkov/crane";
-      inputs.nixpkgs.follows = "nixpkgs";
-    };
+    crane.url = "github:ipetkov/crane";
     rust-overlay = {
       url = "github:oxalica/rust-overlay";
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -2,10 +2,7 @@
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
     flake-utils.url = "github:numtide/flake-utils";
-    crane = {
-      url = "github:ipetkov/crane";
-      inputs.nixpkgs.follows = "nixpkgs";
-    };
+    crane.url = "github:ipetkov/crane";
```
tomkarw, line 5: LGTM. I was skeptical about removing `crane.inputs.nixpkgs.follows = "nixpkgs";`, but apparently they no longer rely on `nixpkgs` directly (that's what the warning/update was about, duh).

tomkarw (approved):
LGMT!

CI is failing due to some unrelated check failing, I'll update Cargo dependencies and it will make this pipeline green.

flake.nix
ulucs: Yes, they no longer pin the version; so following is no longer necessary

mara-schulke (approved):

PR #272, opened 2024-11-21
Feature/registry alias package

dgehriger:
This PR ensures that registry aliases are resolved prior to packing / publishing Proto.toml packages. The original Proto.toml is included as _Proto.toml.orig_. This behavior mirrors what Cargo is doing.

In addition, this PR adds a top-level option `--insecure` / `-k` to disable SSL validation. This option is useful for some use cases where `buffrs` runs inside e.g. a Docker container whose root certs aren't able to validate the server's cert.


dgehriger:
Closed in favor of #273

PR #273, opened 2024-11-21
Feature/registry alias

dgehriger:
This PR introduces these features:

This release introduces several key updates and new features:

---

#### **New Command-Line Options**
- **`-k` / `--insecure`**: Skip SSL validation during operations. Useful for environments with self-signed certificates or testing scenarios.

---

#### **Enhancements to `buffrs install` Subcommand**
- **`--buf-yaml` Flag**:  
  Adding this flag during `buffrs install` (e.g., `buffrs install --buf-yaml`) generates a _buf.yaml_ file matching installed dependencies during dependency resolution.  
  - This enables seamless use of `buf` for linting and generating proxies/stubs from proto files.
- **`--only-dependencies` Flag**:
   Only installs vendor dependencies.

---

#### **Default Arguments in Configuration**
- **Specify Defaults via _.buffrs/config.toml_:**  
  Users can now define default arguments for commands. For example, to always generate _buf.yaml_ files during `buffrs install`:

```toml
[registry]
default = "globus"

[registries]
globus = "https://conan-us.globusmedical.com/artifactory"

[commands.install]
default_args = ["--buf-yaml"]
```

---

#### **Support for Registry Aliases**
- **Aliases in _Proto.toml_:**  
  The `registry` field in _Proto.toml_ now supports aliases. For instance:

```toml
edition = "0.9"

[package]
type = "api"
name = "grpc-test"
version = "0.1.0"

[dependencies.api-examples-hub]
version = "=0.2.0"
repository = "grpc"
registry = "acme" # Alias for "https://conan-us.acme.com/artifactory"
```

---

#### **Normalized Proto.toml for Compatibility**
- **Registry Aliases in Packaging and Publishing:**  
  Buffrs resolves registry aliases when packing or publishing to an upstream repository. To ensure compatibility, the original _Proto.toml_ is preserved as _Proto.toml.orig_, while the normalized version is used during operations.  
  This mirrors Cargo's behavior, providing a seamless user experience.

##### **Example: Alias Normalization**
**Original _Proto.toml_:**
```toml
edition = "0.9"

[package]
type = "api"
name = "api-examples-test"
version = "0.3.0"

[dependencies.lib-geom]
version = "=0.1.0"
repository = "grpc"
registry = "acme"
```

**Generated _Proto.toml_:**
```toml
# THIS FILE IS AUTOMATICALLY GENERATED BY BUFFRS
#
# Buffrs normalizes Proto.toml files for compatibility with all versions.
# It rewrites `path` dependencies to registry dependencies.
#
# For the original contents, refer to Proto.toml.orig.

edition = "0.9"

[package]
type = "api"
name = "api-examples-test"
version = "0.3.0"

[dependencies.lib-geom]
version = "=0.1.0"
repository = "grpc"
registry = "https://conan-us.acme.com/artifactory"
```


src/config.rs
```diff
@@ -21,9 +21,9 @@ use std::{
 };
 
 // Location of the configuration file
-const CONFIG_FILE: &str = ".buffrs/config.toml";
```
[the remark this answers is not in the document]
dgehriger: @mara-schulke : thank you. Yes, I had been researching the same, and could find plenty of evidence for _$HOME/.config/**_, but none for _<repo>/.config/_. Hence my push-back.

I know it's a big PR, but I'm currently testing that code in a full workflow. I'll keep you posted when it is ready!

src/credentials.rs
```diff
@@ -32,7 +31,7 @@ pub const CREDENTIALS_FILE: &str = "credentials.toml";
 #[derive(Debug, Default, Clone)]
 pub struct Credentials {
     /// A mapping from registry URIs to their corresponding tokens
-    pub registry_tokens: HashMap<RegistryUri, String>,
```
dgehriger, line 35: Note: a `RegistryUri` may now be an actual `Uri`, but also an alias (resolved or not). In the rare case in this code where a URI cannot (longer) be an alias, but must be an actual URI, I have changed the type to `Uri` to enforce this guarantee.

src/main.rs
dgehriger, line 1: Note: moved most code to _cli.rs_

dgehriger:
@mara-schulke : some of the unit tests are failing due to the change in the manifest version (0.9 → 0.10). I'm fixing this now and will update.

mara-schulke:
Hi this is great thank you so much! One note before I'd start a detailed review: The proto rust module is not necessary as this is covered by Tonic itself:

https://docs.rs/tonic-build/latest/tonic_build/struct.Config.html#method.include_file

Could you maybe revert that command?

Once done I'm happy to review and merge this asap!

src/main.rs
```diff
@@ -12,147 +12,9 @@
...
-use buffrs::{manifest::MANIFEST_FILE, package::PackageType};
-use clap::{Parser, Subcommand};
-use miette::{miette, WrapErr};
-use semver::Version;
+use std::env;
 
-#[derive(Parser)]
-#[command(author, version, about, long_about)]
-#[command(propagate_version = true)]
-struct Cli {
```
mara-schulke, line 27: Please keep the clap related code out of the lib.rs if possible; you can if you find this way of doing it unpleasant use binaries with modules (src/bin/cli/*) but in my personal opinion I'd vouch for as little outsourcing as possible at this point 

But inlining it into the library code is mixing the CLI layer with the actual impl code

src/package/compressed.rs
```diff
@@ -47,76 +47,112 @@ impl Package {
...
+                "failed to create package, manifest doesn't contain a package declaration"
             ));
         }
 
         let mut archive = tar::Builder::new(Vec::new());
 
+        // Add original and resolved manifests
+        Self::add_manifest_to_archive(
+            &mut archive,
+            &manifest.for_publishing()?,
```
mara-schulke, line 67:
```suggestion
            &manifest.resolve()?,
```

I would (maybe) consider adding another manifest new type which holds resolved manifests only and is only constructable by the above fn; that way you can differentiate by the state of the manifest (ie. containing aliases and resolved)

src/package/compressed.rs
```diff
@@ -47,76 +47,112 @@ impl Package {
...
+        let tgz = Self::compress_tarball(tar)?;
+
+        Ok(Self { manifest, tgz })
+    }
+
+    /// Helper to add a manifest (original or resolved) to the tarball.
+    fn add_manifest_to_archive(
+        archive: &mut tar::Builder<Vec<u8>>,
+        manifest: &Manifest,
+        prefix: Option<&str>,
```
mara-schulke, line 106: Please work with path slices here; instead of pushing strs

mara-schulke: That way you can also skip the prefix arg as it can be done by concating path slices on the call site

src/package/compressed.rs
```diff
@@ -47,76 +47,112 @@ impl Package {
...
-            .into_inner()
+            .append_data(&mut header, file_name, Cursor::new(manifest_bytes))
             .into_diagnostic()
-            .wrap_err(miette!("failed to assemble tar package"))?;
+            .wrap_err(miette!("failed to add manifest `{file_name}` to release"))?;
+        Ok(())
+    }
 
+    /// Helper to compress the tarball into a `.tgz` file.
+    fn compress_tarball(tar: Vec<u8>) -> miette::Result<Bytes> {
```
mara-schulke, line 145: Can we make a `Tarball` newtype for this?

mara-schulke: That yields to better separation and easier testability + makes it reusable for other parts of the code

src/package/compressed.rs
```diff
@@ -186,11 +222,11 @@ impl Package {
             .into_diagnostic()
             .wrap_err(DeserializationError(ManagedFile::Manifest))?;
 
-        let manifest = String::from_utf8(manifest)
+        let manifest_str = String::from_utf8(manifest)
```
mara-schulke, line 225: If you disagree with the original way of reassignment I would suggest nesting a block for decoding so you drop the str variant as soon as the block ends

src/package/compressed.rs
```diff
@@ -244,3 +280,15 @@ impl TryFrom<Bytes> for Package {
         Package::parse(tgz)
     }
 }
+
+const MANIFEST_PREFIX: &str = r#"# THIS FILE IS AUTOMATICALLY GENERATED BY BUFFRS
```
mara-schulke, line 284: I would put this into the `ResolvedManifest` newtype!

src/package/store.rs
```diff
@@ -39,7 +41,7 @@ impl PackageStore {
     /// Path to the dependency store
     pub const PROTO_VENDOR_PATH: &'static str = "proto/vendor";
 
-    fn new(root: PathBuf) -> Self {
+    pub(crate) fn new(root: PathBuf) -> Self {
```
mara-schulke, line 44: Why is this needed?

src/package/store.rs
```diff
@@ -150,10 +158,38 @@ impl PackageStore {
...
+    /// - `config` - Configuration to use (for alias resolution)
+    /// - `deps` - Optional dependency graph to fetch dependencies from
+    ///
+    /// # Returns
+    /// A `Package` instance representing the packaged release
+    pub async fn release(
+        &self,
+        manifest: &Manifest,
+        config: &Config,
+        deps: Option<&DependencyGraph>,
```
mara-schulke, line 178: Can you quickly elaborate why this is needed?

src/registry/artifactory.rs
```diff
@@ -24,24 +23,50 @@ use semver::Version;
...
-    pub fn new(registry: RegistryUri, credentials: &Credentials) -> miette::Result<Self> {
+    ///
+    /// # Arguments
+    /// * `registry` - The registry URI
+    /// * `credentials` - The credentials to use for the registry
+    /// * `cert_validation_policy` - The policy for validating artifactory server certificates
+    pub fn new(
+        registry: url::Url,
+        credentials: &Credentials,
+        cert_validation_policy: CertValidationPolicy,
```
mara-schulke, line 55:
```suggestion
        policy: CertValidationPolicy,
```

Nit! Personal opinion: the type name conveys enough info here

src/registry/artifactory.rs
```diff
@@ -58,10 +83,10 @@ impl Artifactory {
     /// Pings artifactory to ensure registry access is working
     pub async fn ping(&self) -> miette::Result<()> {
         let repositories_url: Url = {
-            let mut uri = self.registry.to_owned();
+            let mut uri: url::Url = self.registry.clone();
```
mara-schulke, line 86: See above comment for the enum approach

src/registry/artifactory.rs
```diff
@@ -24,24 +23,50 @@ use semver::Version;
...
+
+    /// Do not validate the certificate
+    NoValidation,
+}
+
 /// The registry implementation for artifactory
 #[derive(Debug, Clone)]
 pub struct Artifactory {
-    registry: RegistryUri,
+    registry: url::Url,
```
mara-schulke, line 40: Please make a newtype for holding only valid Artifactory URIs

src/registry/mod.rs
```diff
@@ -14,68 +14,151 @@
...
+        match self {
+            RegistryUri::Alias(alias) => match config {
+                Some(config) => {
+                    let url = config.lookup_registry(alias)?;
+                    Ok(RegistryUri::ResolvedAlias {
+                        alias: alias.clone(),
+                        url: url.clone(),
+                    })
+                }
+                None => Err(miette!("no configuration provided to resolve alias")),
```
mara-schulke, line 64: Please give the user the information on what alias is affected

src/registry/mod.rs
```diff
@@ -14,68 +14,151 @@
...
+    fn deserialize<D>(deserializer: D) -> Result<RegistryUri, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let url = String::deserialize(deserializer)?;
+        RegistryUri::from_str(&url).map_err(serde::de::Error::custom)
     }
 }
 
-impl Deref for RegistryUri {
```
mara-schulke, line 44: Thanks for removing this; this was an anti pattern

mara-schulke:
You should not invoke the CLI level code in a build.rs as doing network requests as part of your build process is going to yield you flaky builds

I'm not keen on re introducing this pattern (this was actually once possible with buffrs - it turned out to be causing all sorts of issues). Scheduling the installation step of your protos is better done by using (or better integrating buffrs into) a proper build system for your project like nix, bazel etc.

We have done this at Helsing and it works great, incl caching etc.

src/main.rs
dgehriger: @mara-schulke : So my issue is that I have a "build.rs" that wants to call `buffrs`, but without downloading its binary. Moving the CLI logic into the library here allows this to be implemented easily, by simply calling the `run()` function.

I can address your concern in three ways, let me know which one is acceptable:

1. Keep the current solution, and accept that the CLI is part of the lib
2. Split buffrs into a workspace, with a core lib crate, a cli lib crate, and the binary crate. 
3. Move the CLI back into the bin crate, and force third-party build.rs to always go through executing `buffrs` as a binary,

dgehriger: Just read your final comment below. I see. I'll go with the 3rd option for now until we move to bazel ourselves. Or what build system would you recommend ?

mara-schulke: I can wholeheartedly recommend you to look into nix for build orchestration; buffrs already has a nix integration so you can download and build all your protos with buffrs using nix today;

You can absolutely go with the 3rd option for your use case but my above point is that that is also not a desirable end state (none of the three are) but it may be a good compromise for backwards compat.

So how we do it is (using nix)

1. Download protos using the buffrs nix lib
2. Run build of project using cargos nix lib

In your build.rs you would then only call out to `tonic_build` as the protos are already present in the file system.

Both step 1 and 2 can be executed with a single nix build command!

src/registry/artifactory.rs
dgehriger: 52fe4ef5f2fddf2cba769cfe296609f7ec27800a: Actually restored `RegistryUri` and introduced `RegistryRef`

src/registry/mod.rs
dgehriger: 24b728d737a1807a6d09af46915a2829b6a9335b

src/registry/artifactory.rs
dgehriger: 52fe4ef5f2fddf2cba769cfe296609f7ec27800a

src/registry/artifactory.rs
dgehriger: a6a41b4a48640adb766999e3c145db5895e84059

src/package/store.rs
dgehriger: A local dependency may transitively depend on local or remote dependencies. In [`DependencyGraphBuilder::process_local_dependency()`](https://github.com/helsing-ai/buffrs/pull/273/files#diff-cbb50fec325653fd70e0bcb9f84525b95fc31231915716c70457695941f3951dR301), prior to processing local / remote dependencies, a first pass is performed to build up the `DependencyGraph` of transitive (sub-) dependencies.

This is passed into `Store::release()` to assist resolution.

Without this, if a local dependency doesn't have its own (transitive) dependencies already installed (in its own _vendor/_ directory), `Store::release()` would fail.

src/package/store.rs
dgehriger: Left-over. Reverted in cc82f342bff0e9f93a3168b0cba29001e5f410d1.

dgehriger: Actually no. It's needed by the unit test in _buf_yaml.rs_, line 274. I'm going to restore it.

src/package/compressed.rs
dgehriger: I instead introduced a trait `PublishableManifest` and the `ResolvedManifest` newtype. Is that ok for you, too? f77d2faf2d04a4765aace473cea58f2ed5a09cd0

src/package/compressed.rs
dgehriger: f77d2faf2d04a4765aace473cea58f2ed5a09cd0

src/package/compressed.rs
dgehriger: f77d2faf2d04a4765aace473cea58f2ed5a09cd0

src/main.rs
dgehriger: 68d5f43ea1c48bc79dbf42eea3eca62dfcfd055e

dgehriger:
> One note before I'd start a detailed review: The proto rust module is not necessary as this is covered by Tonic itself:
> 
> [docs.rs/tonic-build/latest/tonic_build/struct.Config.html#method.include_file](https://docs.rs/tonic-build/latest/tonic_build/struct.Config.html#method.include_file)
> 
> Could you maybe revert that command?

What a joke... I wasn't aware of this. Of course.

src/package/compressed.rs
dgehriger: Sure: 26395f5ab5ba00862b52918c9d3326ae35fb1a1f. But did you have more in mind that this commit?

src/package/compressed.rs
dgehriger: f77d2faf2d04a4765aace473cea58f2ed5a09cd0

src/config.rs
```diff
@@ -0,0 +1,376 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Config {
+    /// Edition of this configuration file (in sync with the Proto.toml edition)
+    edition: Edition,
+
+    /// Path to the configuration file
+    config_path: Option<PathBuf>,
+
+    /// Default registry alias to use if none is specified
+    default_registry: Option<String>,
```
mara-schulke, line 61: Last new type request: Could we use a registry name new type to ensure that this string is infact a valid registry name?

I would suggest using a similar convention as with package names

mara-schulke: You could basically derive / share a lot of logic with the PackageName type

src/config.rs
```diff
@@ -0,0 +1,376 @@
...
+    /// * `registry` - The registry argument to parse
+    ///
+    /// # Returns
+    /// URI with either alias scheme or actual URI:
+    /// - <alias> -> alias://<alias>
+    /// - <uri> -> <uri>
+    /// - None -> alias://<default>
+    pub fn parse_registry_arg(&self, registry: &Option<String>) -> miette::Result<RegistryRef> {
+        match registry {
+            Some(registry) => RegistryRef::from_str(registry),
```
mara-schulke, line 100: Nit: .parse is more idiomatic imo

src/config.rs
```diff
@@ -0,0 +1,376 @@
...
+
+    /// List of registries
+    registries: HashMap<String, RegistryUri>,
+
+    /// Default arguments for commands
+    command_defaults: HashMap<String, Vec<String>>,
+}
+
+impl Config {
+    /// Create a new configuration with default values
```
mara-schulke, line 71: This comment lies 

Could you roughly follow the public api semantics from the manifest / cache?

src/config.rs
```diff
@@ -0,0 +1,376 @@
...
+    }
+
+    /// Parse a registry argument
+    ///
+    /// # Arguments
+    /// * `registry` - The registry argument to parse
+    ///
+    /// # Returns
+    /// URI with either alias scheme or actual URI:
+    /// - <alias> -> alias://<alias>
```
mara-schulke, line 95: I would, if you still have time for this, ask you to not encode the fact we talk about an alias using the uri scheme if possible

Internally when holding the registry ref we should utilize an enum (as you do) and during serialization we can just allow serializing (or better de ser) we can make the decision depending on whether the registry contains a scheme at all

src/config.rs
```diff
@@ -0,0 +1,376 @@
...
+        }
+    }
+
+    /// Lookup a registry by name
+    ///
+    /// # Arguments
+    /// * `name` - Name of the registry to lookup
+    ///
+    /// # Returns
+    /// The registry URI
```
mara-schulke, line 114: Would you eventually remove the doc comments argument and return type bodies? They will get out of sync and rust doc does a great job at this

src/config.rs
```diff
@@ -0,0 +1,376 @@
...
+/// default = "some_org"
+///
+/// [commands]
+/// default_args = ["--insecure"]
+///
+/// [commands.install]
+/// default_args = ["--generate-buf-yaml"]
+/// ```
+///
+#[derive(Debug, Clone, PartialEq, Eq)]
```
mara-schulke, line 52: Please make this implement Serde as per convention instead of pulling out values manually from the toml value

src/config.rs
```diff
@@ -0,0 +1,376 @@
...
+
+    /// Load registries from the [registries] section of the config file
+    ///
+    /// # Arguments
+    /// * `config` - The parsed TOML configuration
+    /// * `config_path` - Path to the configuration file (for error messages)
+    ///
+    /// # Returns
+    /// A map of registry names to URIs
+    fn get_registries(
```
mara-schulke, line 249: (Related to the serde comment)

You can save a ton of code and error surface + following idiomatic language patterns by doing a quick serde impl here. For me this would be the last blocker to merging this though!

src/integration/buf_yaml.rs
```diff
@@ -0,0 +1,352 @@
...
+
+    vendor_modules.sort();
+    buf_yaml.clear_modules();
+
+    if manifest.package.is_some() {
+        // double-check that the package really contains proto files
+        // under proto/** (but not under proto/vendor/**)
+        let vendor_path = store.proto_vendor_path();
+        let mut has_protos = false;
+        for entry in WalkDir::new(store.proto_path()).into_iter().flatten() {
```
mara-schulke, line 137: Could you utilize PackageStore::list?

src/main.rs
```diff
@@ -281,3 +365,51 @@ fn infer_package_type(lib: bool, api: bool) -> Option<PackageType> {
...
+
+/// Retrieve and merge default arguments with user-provided arguments
+///
+/// # Arguments
+/// * `config` - The configuration object
+/// * `args` - The user-provided arguments
+///
+/// # Returns
+/// A vector of arguments with default arguments merged in
+pub fn merge_args_with_defaults(config: &Config, args: &[String]) -> Vec<String> {
```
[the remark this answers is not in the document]
mara-schulke: Ie. this kind of clippy wrangling is indicating a unintended use of the library whereas default arguments can be trivially solved by shell configuration / build scripts 

Ie. user value vs implementation overhead don't seem to add up here

src/manifest.rs
```diff
@@ -289,14 +301,76 @@ impl Manifest {
...
+                            registry: remote_manifest.registry.with_alias_resolved(config)?,
+                        })
+                    }
+                    DependencyManifest::Local(local_manifest) => {
+                        // For local dependencies, check if a remote manifest is present
+                        // and resolve its registry alias
+                        if let Some(ref remote_manifest) = local_manifest.publish {
+                            DependencyManifest::Local(LocalDependencyManifest {
+                                path: local_manifest.path.clone(),
+                                publish: Some(RemoteDependencyManifest {
```
mara-schulke, line 340: Ie. all of this can be either done in the resolve call or ideally during dependency resolution 

So you would only need to read the config in that part of buffrs instead of everywhere

src/manifest.rs
```diff
@@ -342,32 +412,103 @@ impl Manifest {
...
-        let dependencies = raw
-            .dependencies()
-            .iter()
-            .map(|(package, manifest)| Dependency {
-                package: package.to_owned(),
-                manifest: manifest.to_owned(),
-            })
-            .collect();
+    /// Tests if the manifest is fully resolved (only contains remote dependencies)
+    pub fn assert_fully_resolved(&self) -> miette::Result<()> {
```
mara-schulke, line 417: Please don't make this a public api call

src/registry/mod.rs
```diff
@@ -14,44 +14,161 @@
...
+    }
+}
+
+/// A reference to a registry
+#[derive(Debug, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub enum RegistryRef {
+    /// A URL to a registry
+    Url(RegistryUri),
+    /// An alias to a registry
+    Alias(String),
```
mara-schulke, line 55: In my understand you have two types:

RegistryRef (a reference, potentially aliased registry) and RegistryUri the uri version, independent of its origin

src/registry/mod.rs
```diff
@@ -14,44 +14,161 @@
...
+    /// A resolved alias to a registry
+    ResolvedAlias {
```
mara-schulke, line 57: Thus this case seems both type unsafe and unnecessary to me as you would want to go from RegistryRef to RegistryUri during alias resolution

src/registry/mod.rs
```diff
@@ -14,44 +14,161 @@
...
+        /// The resolved URL
+        url: RegistryUri,
+    },
+}
+
+impl RegistryRef {
+    /// Get the raw URL of the registry with any alias resolved
+    ///
+    /// # Arguments
+    /// * `config` - The configuration to use to resolve the alias
```
mara-schulke, line 69: Ie this could be typesafe!

src/registry/mod.rs
```diff
@@ -14,44 +14,161 @@
...
+                    "no configuration provided to resolve alias \"{}\"",
+                    alias
+                )),
+            },
+            _ => Ok(self.clone()),
+        }
+    }
+
+    /// Serializer for resolved RegistryUris
+    pub fn serialize_resolved<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
```
mara-schulke, line 90: This aswell.

src/registry/mod.rs
```diff
@@ -75,7 +192,16 @@ impl FromStr for RegistryUri {
...
+/// Ensure that the URL is valid for a registry
+///
+/// A valid registry URL must:
+/// - Have a scheme of either "http" or "https"
+/// - End with "/artifactory" if the host is a JFrog Artifactory instance
+/// - Have a host component
+///
+/// # Arguments
+/// * `url` - The URL to check
+pub fn sanity_check_url(url: &Url) -> miette::Result<()> {
```
mara-schulke, line 204: Please don't make this part of the public api

mara-schulke (changes requested):
This is definitely going in the right direction although I feel like this is loosening the type safety and semantic guarantees we currently provide. I'm happy to have another chat about this to get it over the line!

src/command.rs
```diff
@@ -476,18 +540,19 @@ pub async fn list() -> miette::Result<()> {
         let rel = proto
             .strip_prefix(&cwd)
             .into_diagnostic()
-            .wrap_err(miette!("failed to transform protobuf path"))?;
+            .wrap_err(miette!("failed to transform protobuf path"))?
+            .to_posix_string();
 
-        print!("{} ", rel.display())
```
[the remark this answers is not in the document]
dgehriger: On Windows, the paths will use back-slashes. The canonicalized Proto.toml should be the same, no matter if published from Windows or Linux.

src/config.rs
dgehriger: I'll try -- but it seems that I'd have to wrestle Serde to do so.

src/config.rs
dgehriger: Can you give me specifics?

src/config.rs
dgehriger: Can we add this in a folllow-up PR? In fact, I am running out of time with this, so I'd highly appreciate if we could stack some of the requests.

src/config.rs
dgehriger: If you wish. But rustdoc won't tell me that `name` is a registry name to look up. A newtype of course will.

src/integration/path_util.rs
```diff
@@ -0,0 +1,97 @@
+use std::path::{Component, Path, PathBuf};
+
+/// Extension trait for `Path`.
+pub trait PathUtil {
```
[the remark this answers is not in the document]
dgehriger: I'll make that private, agreed. But this is definitely not std. It used to be, but no longer is.

There are third-party crates doing similar things, but they would still require a wrapper function, so I'd argue that this is cleaner. Your call.

src/registry/mod.rs
dgehriger: You mean we could express the Alias as `Alias(String, Option<RegistryUri>)` ?

I found it safer to not use the option, but a third type. But for aliases, we need to remember the alias name and URL at all times, as, depending on the context, we do need to be able to access the alias, even if resolved.

What could be done is to split this enum into two types:

```rust
enum RawRegistryRef {
    Url(RegistryUri),
    Alias(String),
};

enum RegistryRef {
    Url(RegistryUri),
    Alias(String, RegistryUri),
};
```

src/command.rs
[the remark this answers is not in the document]
mara-schulke: Ah nice I see okay! Thanks for this compat effort!

Would you mind making this private then?

src/config.rs
```diff
@@ -0,0 +1,376 @@
...
+/// some_org = "https://artifactory.example.com/artifactory/some-org"
+///
+/// [registry]
+/// default = "some_org"
+///
+/// [commands]
+/// default_args = ["--insecure"]
+///
+/// [commands.install]
+/// default_args = ["--generate-buf-yaml"]
```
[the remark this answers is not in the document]
mara-schulke: Yes that's fine, happy to merge it

src/config.rs
mara-schulke: Ah yes sorry.

So this function will do IO although the comment says it will create a new one with default values!

With regards to the common pattern I was referring to the `Manifest` struct for example

src/config.rs
mara-schulke: You can take a look at the RawManifest!

src/registry/mod.rs
mara-schulke: Well I'm talking about:

```rust
enum RegistryRef {
    Uri(Uri)
    Alias(RegistryName)
}

impl RegistryRef {
    fn resolve(config: &Config) -> Result<RegistryUri> {
        // 
    }
}

struct RegistryUri(Uri);
```

mara-schulke: Why do you need to remember aliases here?

Like you can do the resolving as late as possible (and still keep the original registry ref around)

mara-schulke:
Okay, let's cut any refactorings or feature changes to get it merged asap. If you could tackle the code quality topics I've raised I would be glad. Once you are happy please give me another go and I'll have a look!

src/registry/mod.rs
dgehriger: @mara-schulke : because resolution requires access to the config, and if postponing it, one has to pass the config reference into code that definitely shouldn't have to know about it.

src/config.rs
dgehriger: 0f35c69d2b211fca144ae8cbcec13344b9a4200e

src/config.rs
dgehriger: e48abf21eab4565706ab4f078a2a38f14af05f3f

src/config.rs
dgehriger: e48abf21eab4565706ab4f078a2a38f14af05f3f

src/config.rs
dgehriger: 24136669c705663dd0845c21ca7e77259d55eff8

src/registry/mod.rs
dgehriger: a2dc4e00eda024debb7d9bebe0ebb7ece03287f7

src/registry/mod.rs
dgehriger: Let's push changing this to a follow-up PR, ok?

PR #274, opened 2024-12-06
Keeping files "last modified" time to enable change detection in build.rs

vsiles:
Most codegen tools (list prost or tonic) are trying to detect if the proto files changed (with things like `println!("cargo:rerun-if-changed={}", proto);`), which is based on the last modification time of a file.

In the current situation, all files added to a buffrs `Package` are stripped from their modification time (it is set to EPOCH by `tar`), so `build.rs` like scripts can't detect if a file changed or not after calling `buffrs install`.

In this PR, I keep the mtime around to fix this situation and enable builder to only to the codegen when necessary.

Some design decisions:
- Do we want this for the manifest files too ? For now, only the proto files have their mtime kept around
- Do we want an option to `buffrs package` to select if we preserve or not the mtime ?
- If we want such an option, what default should we use (I vote to "preserve" by default, like tar: https://docs.rs/tar/latest/tar/struct.Archive.html#method.set_preserve_mtime)

### DIFF
--- a/src/package/store.rs
+++ b/src/package/store.rs
@@ -152,7 +152,11 @@ impl PackageStore {
     }
 
     /// Packages a release from the local file system state
-    pub async fn release(&self, manifest: &Manifest) -> miette::Result<Package> {
+    pub async fn release(
+        &self,
+        manifest: &Manifest,
+        preserve_mtime: bool,
+    ) -> miette::Result<Package> {
         for dependency in manifest.dependencies.iter() {
             let resolved = self.resolve(&dependency.package).await?;
 
@@ -182,7 +186,7 @@ impl PackageStore {
             );
         }
 
-        let package = Package::create(manifest.clone(), entries)?;
+        let package = Package::create(manifest.clone(), entries, preserve_mtime)?;
 
         tracing::info!(":: packaged {}@{}", package.name(), package.version());
 
--- a/src/package/store.rs
+++ b/src/package/store.rs
@@ -16,7 +16,6 @@ use std::{
     collections::BTreeMap,
     env::current_dir,
     path::{Path, PathBuf},
-    time::UNIX_EPOCH,
 };
 
 use bytes::Bytes;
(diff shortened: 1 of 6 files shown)


mara-schulke:
Im pro making this a flag of the package / publish commands

src/package/store.rs
```diff
@@ -260,6 +275,13 @@ impl PackageStore {
     }
 }
 
+pub struct Entry {
+    /// Actual bytes of the file
+    pub contents: Bytes,
+    /// last modified time, as number of seconds since EPOCH, if available
+    pub mtime: Option<u64>,
```
mara-schulke, line 282: Can we eventually just store the whole metadata structure here? https://doc.rust-lang.org/nightly/std/fs/struct.Metadata.html

mara-schulke (approved):

vsiles:
> Im pro making this a flag of the package / publish commands

What default behavior do you want ? (please they preserve :p)

mara-schulke:
Yea that's fine! We can use the other one (ie resetting them to epoch) explicitly

src/resolver.rs
```diff
@@ -102,13 +102,45 @@ struct DownloadError {
...
+    lockfile: &'a Lockfile,
+    credentials: &'a Arc<Credentials>,
+    cache: &'a Cache,
+    preserve_mtime: bool,
+}
+
+struct ProcessLocalDependency<'a> {
+    name: PackageName,
+    dependency: LocalDependency,
+    #[allow(dead_code)]
```
vsiles, line 118: @mara-schulke refactoring to avoid the "too many function arguments" warning here, and it showed that this input is never read. I left it for you to deal with it as you see fit.

vsiles:
Update:
- adding CLI options to enable (default) or disable it. Notably `install` needs it too since it packages local files too
- refactoring some internal API to avoid "too many function parameters" warning

vsiles:
@mara-schulke clippy is not happy (and doesn't repro locally when I run `cargo`). It seems it's due to some unrelated changes for this MR. Should I ignore/bypass once you reviewed the MR, or should we make a separate MR to clean that up ?
Any idea how to repro locally ?

vsiles:
CI should now be happy ;)

PR #276, opened 2025-04-09
Using generated code for manifest serializing/deserializing instead of custom parsers

P-E-Meunier:
This is just simple code cleanup, and will allow to add more formats to the manifest.


src/manifest.rs
```diff
@@ -125,105 +151,44 @@ impl RawManifest {
...
-                    ref dependencies,
-                } => {
-                    let mut s = serializer.serialize_struct("Unknown", 2)?;
-                    s.serialize_field("package", package)?;
-                    s.serialize_field("dependencies", dependencies)?;
-                    s.end()
-                }
-            }
+#[test]
+fn manifest() {
```
mara-schulke, line 155: Could you please add tests that:

1. Verify that the default edition is `CANARY` during serialization
2. Serialization fails when the edition doesnt match the format specified?

Also (nit) could you group them in a test mod?

src/manifest.rs
```diff
@@ -86,13 +99,26 @@ impl From<Edition> for &'static str {
...
 /// A buffrs manifest format used for serialization and deserialization.
 ///
 /// This contains the exact structure of the `Proto.toml` and skips
 /// empty fields.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(untagged)]
 enum RawManifest {
     Canary {
+        edition: Edition,
```
mara-schulke, line 121: This will most likely break stuff.

The old deserialization code mapped multiple different versions to the `RawManifest::Canary` format. In the future this is great to maintain backwards compat: Ie. you can introduce a new format of the `RawManifest` (e.g. `V1`) and determine into which one to deserialize by looking at the edition field first.

This MR lost that capability as far as i can tell: You deserialize everything that has an edition into `Canary` and everything without into `Unknown` (which is indeed the current behaviour, but rules out easy adoption of breaking manifest changes). If you now add the third variant to your code, you are not inspecting the edition format pre decoding.

One option i could see to recover this (if you feel strongly about not having manual deserialization in place) is to add newtypes around `Edition` which represents the subsets supported by one `RawManifest` variant.

mara-schulke (changes requested):
Happy to change / get rid of the manual serialization / deserialization code! Thanks for raising this. My main concerns evolve around the constraints wrt. edition management! If we can bring back the "This edition must serialize into this raw format" then im happy to roll with this!

PR #277, opened 2025-04-09
Ensure stable order of print_files output

adamdayan:
buffrs lock print-files is used to generate paths for the nix buffrs-cache derivation. To ensure this derivation doesn't change when the underlying Proto.lock has not changed, print-files must output files in a stable order. To ensure this, replace Lockfile's unordered HashMap with an ordered BTreeMap

### DIFF
--- a/src/lock.rs
+++ b/src/lock.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::HashMap;
+use std::collections::BTreeMap;
 
 use miette::{ensure, Context, IntoDiagnostic};
 use semver::Version;
@@ -135,7 +135,7 @@ struct RawLockfile {
 /// Used to ensure future installations will deterministically select the exact same packages.
 #[derive(Default)]
 pub struct Lockfile {
-    packages: HashMap<PackageName, LockedPackage>,
+    packages: BTreeMap<PackageName, LockedPackage>,
 }
 
 impl Lockfile {
@@ -221,6 +221,10 @@ impl FromIterator<LockedPackage> for Lockfile {
 }
 
 impl From<Lockfile> for Vec<FileRequirement> {
+    /// Converts lockfile into list of required files
+    ///
+    /// Must return files with a stable order to ensure identical lockfiles lead to identical
+    /// buffrs-cache nix derivations
     fn from(lock: Lockfile) -> Self {
         lock.packages.values().map(FileRequirement::from).collect()
     }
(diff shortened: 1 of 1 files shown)


src/lock.rs
```diff
@@ -293,3 +294,60 @@ impl From<&LockedPackage> for FileRequirement {
...
+
+    use crate::{package::PackageName, registry::RegistryUri};
+
+    use super::{Digest, DigestAlgorithm, FileRequirement, LockedPackage, Lockfile};
+
+    #[test]
+    fn stable_file_requirement_order() {
+        let lock = Lockfile {
+            packages: BTreeMap::from([
+                (
```
mara-schulke, line 312: I would add more packages here to increase the chance of a potential wrong ordering occurring

src/lock.rs
```diff
@@ -293,3 +294,60 @@ impl From<&LockedPackage> for FileRequirement {
...
+                        dependencies: vec![],
+                        dependants: 1,
+                    },
+                ),
+            ]),
+        };
+
+        let files: Vec<FileRequirement> = lock.clone().into();
+        for _ in 0..30 {
+            let other_files: Vec<FileRequirement> = lock.clone().into();
```
mara-schulke, line 349: Im not sure that you are actually testing anything here. You are cloning which is a usually byte by byte memcpy - I believe you would persist any initial random ordering occurred from building a HashMap.

https://docs.rs/hashbrown/latest/src/hashbrown/map.rs.html#190-204

https://docs.rs/hashbrown/latest/src/hashbrown/raw/mod.rs.html#3238

I can't verify right now, but does this test fail if you plug in a HashMap instead of a BTreeMap? 

I think testing the ordering a little more explicit here would be good! Maybe using two different instances of the struct instead of cloning it (which would also get rid of the cfg_attr attribute)

mara-schulke (changes requested):

src/lock.rs
```diff
@@ -134,8 +134,9 @@ struct RawLockfile {
 ///
 /// Used to ensure future installations will deterministically select the exact same packages.
```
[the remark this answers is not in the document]
adamdayan: I've added a comment to that effect on `From<Lockfile> for Vec<FileRequirement>` because the locked packages don't necessarily need to be stored in order, just turned into a Vec in order

src/lock.rs
adamdayan: done

adamdayan:
there seem to be a bunch of clippy lints and security advisories failing in CI on my MR that aren't related to my changes - should I fix them in this MR, fix them in a different MR or are they not required to be fixed before merging this?

vsiles:
> there seem to be a bunch of clippy lints and security advisories failing in CI on my MR that aren't related to my changes - should I fix them in this MR, fix them in a different MR or are they not required to be fixed before merging this?

I would fix them in another MR and rebase this one on top of it. Do you have the time to do it ?

mara-schulke (approved):

PR #278, opened 2025-04-10
Bump nix flake Rust input, fix clippy errors and RUSTSEC warnings

adamdayan:
The clippy github action uses rust 1.86.0, which introduced new lints that buffrs currently fails. This MR updates the flake Rust dependency to 1.86.0, fixes the clippy lints and all the new RUSTSEC warnings.


vsiles:
Feels like `paste` is no longer there, you can remove the deny entry.

PR #283, opened 2025-06-20
feat(publish): skip publishing if the package is already there

vsiles:
Publishing the same package twice currently triggers a hard error (usually a 403). If the package is there, let's call it a success

### DIFF
--- a/src/registry/artifactory.rs
+++ b/src/registry/artifactory.rs
@@ -15,6 +15,7 @@
 use super::RegistryUri;
 use crate::{
     credentials::Credentials,
+    lock::DigestAlgorithm,
     manifest::{Dependency, DependencyManifest},
     package::{Package, PackageName},
 };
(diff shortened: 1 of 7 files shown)


vsiles:
Got trouble running the tests locally, lots of 404 like
```
Error:   × failed to publish `lib` to `http://127.0.0.1:32849/registry:my-repository`
  ╰─▶ HTTP status client error (404 Not Found) for url
      (http://127.0.0.1:32849/registry/my-repository/lib/lib-0.0.1.tgz)
```
except on the very first try. Am I missing anything (did `nix develop` and then `cargo test`)

src/registry/artifactory.rs
```diff
@@ -230,6 +230,23 @@ impl Artifactory {
...
+            .status();
+        if status.is_success() {
+            tracing::info!(
+                ":: {}/{}@{} is already published, skipping",
+                repository,
+                package.name(),
+                package.version()
+            );
+            return Ok(());
+        }
```
mara-schulke, line 248: Hi @vsiles this looks non optimal, as we **should error** if the package already exists with a different hash. E.g. if actual package contents differ, id like the user to explicitly allow that via a flag or so.

This is because:

1. Duplicated publishing with the same hash (e.g. `abc`) does no harm – end result is the same – a package with hash `abc` was published to the registry.
2. Duplicated publishing with different hashes (e.g. `xyz1` and `xyz2`) does harm – actual contents diverge and the fact that you are trying to publish two diverging sets of protos under one version is not correct. – end result is that the earlier publishing step (e.g. `xzy2`) was published – if the pipeline doesnt fail, or atleast emit a warning – it will take very long for the user to understand that their protos (`xzy1`) were actually not published.

mara-schulke: I think you can just try to grab the hash from artifactory to compare fwiw!

mara-schulke:
> Got trouble running the tests locally, lots of 404 like
> 
> ```
> Error:   × failed to publish `lib` to `http://127.0.0.1:32849/registry:my-repository`
>   ╰─▶ HTTP status client error (404 Not Found) for url
>       (http://127.0.0.1:32849/registry/my-repository/lib/lib-0.0.1.tgz)
> ```
> 
> except on the very first try. Am I missing anything (did `nix develop` and then `cargo test`)
> 
> EDIT: ha, they are also here in CI. So I must be doing something wrong in the test I added. Any idea what is the problem ?

My hunch would be that the local registry doesn't implement this endpoint properly – although impl looks dead simple and not obviously wrong.. – happy to spend some time on monday with you to debug

vsiles:
> > Got trouble running the tests locally, lots of 404 like
> > ```
> > Error:   × failed to publish `lib` to `http://127.0.0.1:32849/registry:my-repository`
> >   ╰─▶ HTTP status client error (404 Not Found) for url
> >       (http://127.0.0.1:32849/registry/my-repository/lib/lib-0.0.1.tgz)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > except on the very first try. Am I missing anything (did `nix develop` and then `cargo test`)
> > EDIT: ha, they are also here in CI. So I must be doing something wrong in the test I added. Any idea what is the problem ?
> 
> My hunch would be that the local registry doesn't implement this endpoint properly – although impl looks dead simple and not obviously wrong.. – happy to spend some time on monday with you to debug

Ok ! I'll have look sometime next week (feel free to have a look too :D) but pair debugging will be hard: I'm traveling to Spain for the week. This can also wait for the week after next. Let's sync on Friday

vsiles:
@mara-schulke can you get a final look ?

src/registry/artifactory.rs
mara-schulke: Perfect; Thanks!

mara-schulke (approved):

PR #285, opened 2025-10-27
Add Maven registry support

### LINKED ISSUE #284: Maven registry backend
Would you be open to a contribution that adds Maven backend support?

I think the base work here is:
- add a new Maven implementation that's similar to the Artifactory one but uses maven-metadata.xml instead of Artifactory queries
- likely pulling out a Registry trait for both to implement
- plumbing to allow specifying a registry is Artifactory or Maven, maintaining Artifactory as a default

There's some design decision questions here, in particular how the manifest should store the registry type. A couple of options on that front:
- take a crates style approach and embed "maven+" or "artifactory+" in front of a registry URI to encode type (with absence defaulting to artifactory)
- use a nested type that encodes type/URL separately
- whether this deserves a new rev of the manifest or if it should be handled transparently

markelliot:
Fixes #284

Add a Maven registry backend for buffrs:
 - Extract a Registry trait
 - Encode registry type as a prefix in RegistryUri
 - Dispatch to the correct Registry implementation from RegistryUri


src/registry/maven.rs
```diff
@@ -0,0 +1,446 @@
...
+        );
+
+        ensure!(
+            value.status() != 401,
+            "unauthorized - please provide registry credentials with `buffrs login`"
+        );
+
+        value.error_for_status().into_diagnostic().map(Self)
+    }
+}
```
mara-schulke: Might make sense to share `RequestBuilder` and `ValidatedResponse` between artifactory and maven?

mara-schulke:
Overall im very happy with the code quality and the implementation approach of this pr! Thank you! 😊 

My two dots:

1. This PR is missing tests, before merging i'd like to have end to end tests that ensure that: mixed maven and artifactory setups, maven publishing and maven installation are working as expected. You might need to change the e2e test setup slightly to account for having two registries.
2. Left some comments on type organisation and the possibility to use a builder pattern. Would prefer if they get addressed but it's not a blocker to merging if you feel strongly about the current implementation.

src/registry/maven.rs
```diff
@@ -0,0 +1,446 @@
...
+                    return Err(miette!(
+                        "unable to publish {} to Maven: package is already published with a different hash",
+                        package.name()
+                    ));
+                }
+            }
+        }
+
+        // Upload the artifact
+        let _ = self
```
markusz: Two remarks:

- Unused let binding. 
- The way this is written makes it hard to understand that `body()` and `send()` are implemented on `RequestBuilder` and return a `ValidatedResponse`. Since this resembles the reqwest API, my first impulse was to flag the missing `error_for_status()` (which you correctly implement in the `TryFrom<Response>`)

markusz (commented):
Hi Mark,

thanks a lot for your MR. Having Maven support would be awesome.

I can just echo what Mara said: Overall, this is a high quality MR already. I left a few comments (mostly nits).

The main thing would be to make tests work again and add new unit and e2e tests. The test harness of buffrs can be a bit hard to grok at first. I recently wrestled with it myself, so feel free to reach out if you need help.

I also want to point your attention to v0.12.0 which is soon going to be released (see [MR](https://github.com/helsing-ai/buffrs/pull/286)). This change will cause merge conflicts with your current branch. I took the liberty to merge your branch back into a [dedicated branch](https://github.com/helsing-ai/buffrs/tree/mz/feat/add-workspace-support-maven) of v0.12.0 feature branch and resolve the conflicts to the best of my knowledge. This will hopefully make it easier to merge your changes back to main once finished.

Thanks again for your contribution

src/registry/maven.rs
markelliot: fwiw, this is identical to the artifactory.rs impl

markelliot:
@mara-schulke @markusz ready for another look when you have a minute

mara-schulke (approved):

markusz:
Valid point, but I feel we should value internal consistency over backwards compatibility in this case. Introducing a one-time breaking change as discussed above imo is more digestible than having this long term inconsistency in the Proto files.

Another option would be to make this somehow optional (CLI flag, ENV variable, ..) but that opens another can of worms. 

I'm inclined to merge and be very vocal about the change

@markelliot @mara-schulke  WDYT?

mara-schulke:
Thank you, this is a good catch @markusz - I'd argue we open a tracking issue for 1.0 in GitHub and note this as a breaking change to make in there. From my pov this is would be a breaking change without a strong requirement to be one which seems to push friction downstream for no apparent reason beyond minimally cleaner code.

buffrs is for the most part stable and only missing an advanced dependency resolution mechanism (think pubgrub) for us to make it to 1.0, id like to batch this change with other breaking changes towards the 1.0 release:)

Best,
Mara

markelliot:
@mara-schulke @markusz do you want to proceed here?

mara-schulke:
Hi @markelliot, yes of course! Sorry I have missed this. Can you confirm that the default behavior is this to omit the artifactory prefix when writing? I'm happy with merging this:)

markelliot:
That’s not the default behavior, it can be. If I do that cleanup and resolve the new merge conflicts will you be ready to merge then?

markelliot:
@mara-schulke I've rebased and cleaned up the conflicts

markelliot:
also for full disclosure we ended up going down a different path than using buffrs, so also can understand if you'd prefer to close this without merging. I spent the time to clean it up because you had already spent time engaging on reviewing it and it might be a nice buffrs addition anyway.

PR #290, opened 2026-01-07
Add a verbose mode

mara-schulke:
This merge request is introducing a verbose mode to buffrs which allows debug logs to be output with level, src file, location, and a random prefix (to distinguish multiple parallel buffrs invocations from eachother).

Example output without verbose mode (stays as is):

```
:: installed foo-package@0.1.0
```

Example output in verbose mode:

```
[6fed92f2] INFO  src/operations/installer.rs:310 buffrs::operations::installer: installed foo-package@0.1.0
[6fed92f2] INFO  src/operations/installer.rs:155 buffrs::operations::installer: running install for package: bar-package
[6fed92f2] INFO  src/operations/installer.rs:250 buffrs::operations::installer: bar-package@0.1.0
[6fed92f2] DEBUG src/resolver.rs:374 buffrs::resolver: resolved bar-package@0.1.0 from local cache
[6fed92f2] DEBUG src/operations/installer.rs:399 buffrs::operations::installer: using workspace lockfile version for bar-package@0.1.0
[6fed92f2] DEBUG src/package/store.rs:145 buffrs::package::store: unpacked bar-package@0.1.0 into foo-package/proto/vendor/bar-package
```

This allows to debug and fix issues in production builds. Verbose output can also be toggled via `BUFFRS_VERBOSE=true` so that all subsequent buffrs commands output debug info (useful in CI).

### DIFF
--- a/src/validation/data/package.rs
+++ b/src/validation/data/package.rs
@@ -1,3 +1,4 @@
+// NOTE: `PackageError` fields are allowed unused as they are only used for err formatting
 #![allow(unused_assignments)]
 
 // Copyright 2023 Helsing GmbH
--- a/src/validation/data/packages.rs
+++ b/src/validation/data/packages.rs
@@ -1,3 +1,4 @@
+// NOTE: `PackagesError` fields are allowed unused as they are only used for err formatting
 #![allow(unused_assignments)]
 
 // Copyright 2023 Helsing GmbH


src/operations/publisher.rs
```diff
@@ -186,112 +297,306 @@ impl Publisher {
...
+    /// 2. Reads the package manifest (or uses provided manifest)
     /// 3. Replaces local dependencies with their published remote versions
     /// 4. Creates a package with the updated manifest
     /// 5. Publishes to the registry
     /// 6. Records the mapping of local path to remote location
-    async fn publish_package_at_path(&mut self, package_path: &Path) -> miette::Result<()> {
+    async fn publish_package_at_path(
+        &mut self,
+        package_path: &Path,
+        manifest_override: Option<&PackagesManifest>,
```
mara-schulke: Note @markusz this is now handing the overridden package manifest all the way through until publishing which gets rid of the hash mismatch errors

src/validation/data/package.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(unused_assignments)]
```
Tpt: Might be nice to comment why it's useful.

Tpt (approved):

PR #308, opened 2026-04-15
Prevent 'missing package declaration' errors during publishing

mara-schulke:
This PR repairs and prevents 'missing package declaration' errors in the future, as it creates a new type around package manifests, which ensures that a manifest is publishable before handing it to the actual publishing code. This prevents manifests without a `[package]` section being passed onto branches which can't work without this section using type signatures.

### DIFF
--- a/src/manifest/package.rs
+++ b/src/manifest/package.rs
@@ -113,6 +113,43 @@ impl PackagesManifest {
     }
 }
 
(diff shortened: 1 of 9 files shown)


vsiles (approved):
The changes LGTM. I couldn't really check you didn't miss a spot, so please make sure all locations are updated. But core logic is fine !

PR #310, opened 2026-04-17
Add include/exclude fields to package manifest

asmello:
## Summary

- Adds `include` and `exclude` fields to `PackageManifest`, allowing packages to control which files are included when packaging and populating vendor directories
- `include` starts from a blank slate — only files matching the provided globs are included (any file type, not just `.proto`)
- `exclude` starts from all files and removes those matching the provided globs
- The two fields are mutually exclusive — specifying both produces a validation error at manifest parse time
- When neither is set, the existing default behavior applies (all `.proto` files recursively)
- Replaces `walkdir` with the `ignore` crate for glob matching, type filtering, and override support in `PackageStore::collect()`
- Introduces `Canary13` edition to snapshot current behavior before this change

## Test plan
- [x] Unit tests for serialization/deserialization with include, exclude, both, and neither
- [x] Unit tests for `collect()` with include patterns, exclude patterns, vendor filtering, file-only results, and invalid globs
- [x] Integration tests pass (`cargo test`)
- [ ] Manual smoke test: create a package with `include` / `exclude` in `Proto.toml` and confirm correct files are packaged


src/package/store.rs
```diff
@@ -298,29 +307,104 @@ impl PackageStore {
     ///
     /// * `path` - The root directory to search
     /// * `vendored` - If `false`, excludes files from `proto/vendor/` (for collecting source files only), if `true`, includes all `.proto` files regardless of location
-    pub async fn collect(&self, path: &Path, vendored: bool) -> Vec<PathBuf> {
-        let mut paths: Vec<_> = WalkDir::new(path)
-            .into_iter()
-            .filter_map(Result::ok)
-            .map(|entry| entry.into_path())
-            .filter(|path| {
+    pub async fn collect(
```
asmello, line 316: This is the core of the changes

vsiles:
Will have a look today. Before this, can you motivate the choice of `include` allowing non proto file while the existing behavior filters for .proto only ? A real need or just making implementation more straightforward ?

src/package/store.rs
```diff
@@ -298,29 +313,104 @@ impl PackageStore {
...
+        &self,
+        path: &Path,
+        vendored: bool,
+        include: Option<&[String]>,
+        exclude: &[String],
+    ) -> miette::Result<Vec<PathBuf>> {
+        debug_assert!(
+            include.is_none() || exclude.is_empty(),
+            "include and exclude are mutually exclusive"
```
mara-schulke, line 321: Why is this?

mara-schulke: In my naive understanding i would have assumed that we have:

```
Base = ./proto/**.proto
Include = <some-globs>
Exclude = <some-globs>

Final = (Base & Include) - Exclude
```

src/package/store.rs
```diff
@@ -298,29 +313,104 @@ impl PackageStore {
...
+                    Match::Whitelist(_) => false,
+                }
+            });
+        } else {
+            // Default: include only .proto files
+            let proto_types = {
+                let mut types = TypesBuilder::new();
+                types.add("proto", "*.proto").expect("valid name");
+                types.select("proto");
+                types.build().expect("no conflicting definitions")
```
mara-schulke: While this is the case, i would recommend to bubble up the error nonetheless and with an `Internal` type, just to keep the pretty printing and presentation here

src/package/store.rs
```diff
@@ -348,31 +438,44 @@ impl PackageStore {
...
+                .wrap_err_with(|| {
+                    format!(
+                        "failed to remove directory {} and its contents",
+                        target_dir.display()
+                    )
+                })?;
         }
 
-        for entry in self.collect(&source_path, false).await {
+        let include = manifest.include.as_deref();
```
mara-schulke, line 467: Why is the include defined using `let` but the exclude one is passed as an arg directly?

mara-schulke:
Overall i think this is a sane feature, and a non controversial implementation, but im not yet 100% sure on the semantics of the include and exclude fields. How is this implemented in prevalent package managers? I would go with the most common semantics here; if this is already the case, ignore my comments:)

mara-schulke:
[This seems related](https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields)

mara-schulke:
Also, can you in any case, please extend the book for this feature?

wngr:
@vsiles 
> Before this, can you motivate the choice of include allowing non proto file while the existing behavior filters for .proto only ? A real need or just making implementation more straightforward ?

We have a use case where we want to package an additional manifest file alongside the proto files. This additional metadata is related to the proto definitions, but inlining it as comments seems brittle and cumbersome.

src/package/store.rs
asmello: Yeah, that's also possible (and in fact at first I implemented it that way). But that's not how Cargo works, and apparently the reason is that you never really _need_ both, and there's more than one possible interpretation if both are provided (although I agree what you described is the most intuitive), so we avoid ambiguity and potential mistakes by disallowing this case. With only one list, you can directly infer the resulting filter, while with both it gets trickier to reason about what ends up being included.

src/package/store.rs
asmello: Ok, that can be done. I suppose as long as we end up aborting it works out the same.

asmello:
> [This seems related](https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields)

Entirely! I implemented the same semantics except that when omitted only `*.proto` files are included (for backward compatibility).

src/package/store.rs
asmello: This is because exclude is `Vec<_>` while include is `Option<Vec<_>>` — we need to distinguish `include` not being set from it being set as an empty vec because those have different semantics. But now that I think about it, the same goes for exclude, because an empty exclude list actually includes more files than the default. Will update.

asmello: Updated.

src/package/store.rs
asmello: Done.

Issue #205, opened 2023-12-21
Implement dependency specification resolution


qsantos:
https://github.com/helsing-ai/buffrs/pull/18 brought support for the syntax of dependency specification. However, currently, only pinning (e.g. `=X.Y.Z`) is actually implemented. Dynamic dependency resolutions looks like a sought-after features among users to allow updating a dependency without having to update all the intermediary packages.

See also [PubGrub](https://github.com/pubgrub-rs/pubgrub).

heatonmatthew:
This would be a useful feature.  I'm just getting into Rust, but I might try to find time to look into implementing this.  Has there been any design thinking or progress done that I should be aware of as I start to think about it?

heatonmatthew:
The PubGrub crate is Mozilla Public License 2.0.  I presume that wouldn't be a problem with this project's Apache license?

mara-schulke:
Hi @heatonmatthew, there has not been extensive system design of this beyond some concepts and drafts which indicated that this will require significant internal rewiring of buffrs!:)

PR #311, opened 2026-04-18
Support semver version ranges for dependencies

heatonmatthew:
## Summary

This addresses and implements #205 

- **Removes the exact-pin-only restriction** — dependencies can now be specified with any semver range (`^1.0`, `~1.2.0`, `>=1.5.0`, `=1.0.0`, etc.), not just `=x.y.z`
- **Adds `list_versions()` and `resolve_version()` to `Artifactory`** — queries the registry for all available versions and selects the highest one satisfying the requirement; `get_latest_version()` is reimplemented on top of this
- **`download()` now takes an explicit `&Version`** — resolution is decoupled from downloading; both `Artifactory` and `LocalRegistry` receive the resolved version directly
- **Adds `find_satisfying()` helpers to lockfile types** — `PackageLockfile`, `WorkspaceLockfile`, and the `Lockfile` enum can now look up the best locked version matching a `VersionReq`, enabling the installer to skip the registry for already-locked ranges
- **Removes `dependency_version_string()`** from `src/registry/mod.rs` along with the `VersionNotPinned` error — the old pin-enforcement code is gone entirely
- **Documentation** — new reference pages for semver range syntax (`docs/src/reference/semver.md`) and the resolver (`docs/src/reference/resolver.md`); updated `specifying-dependencies.md` and `buffrs-add.md`
- **Tests** — five new workspace-level integration test scenarios: caret/tilde resolution, compatible diamond, incompatible diamond, and multi-level tree; mock registry extended with query-string support to back these tests

## What a reviewer should know

The resolver remains single-pass (interleaves graph walking and downloading). Version resolution now works as:
1. Check the lockfile for a version satisfying the requirement → use it if found
2. Check the local cache for a matching version → use it if found
3. Call `Artifactory::resolve_version()` to query the registry → download the resolved exact version and cache it

Conflict detection (incompatible diamond) is handled in `validate_version_compatibility()` in `src/resolver.rs` — unchanged in structure, now exercised by the new test cases.

The `query` axum feature was enabled in `Cargo.toml` for the test mock registry to parse search query parameters.

## Test plan

- [x] `cargo build` passes (requires `protoc`)
- [x] `cargo test --workspace` passes (requires `protoc` + Git LFS)
- [x] `cargo clippy --all-targets --workspace -- -D warnings -D clippy::all` passes
- [x] `cargo fmt --check --all` passes
- [x] New integration tests: `cargo test --test e2e install` covers the five new range scenarios

### DIFF
--- a/docs/src/reference/specifying-dependencies.md
+++ b/docs/src/reference/specifying-dependencies.md
@@ -1,52 +1,81 @@
 # Specifying Dependencies
 
-Dependencies are declared in the `[dependencies]` section of the `Proto.toml`
-manifest. Each entry maps a dependency package name to a dependency
-specification object.
+Dependencies are declared in the `[dependencies]` section of `Proto.toml`.
+Each entry names the package and provides a version requirement, registry URL,
+and repository name.
 
-## Remote Dependencies
-
-Remote dependencies are downloaded from an Artifactory registry during
-[`buffrs install`](../commands/buffrs-install.md).
+## Inline table syntax
 
 ```toml
-[dependencies]
-my-package = { registry = "https://your.registry/artifactory", repository = "my-repo", version = "1.2.3" }
+[dependencies.my-lib]
+version = "^1.0.0"
+registry = "https://my-registry.example.com"
+repository = "my-repo"
 ```
 
-The `version` field must be an exact semantic version (e.g. `"1.2.3"`).
-Version ranges or operators (`^`, `~`, `<`, `>`) are not currently supported.
+The three required fields for a remote dependency are:
 
-Use [`buffrs add`](../commands/buffrs-add.md) to add a remote dependency from
-the command line:
+| Field | Description |
+|-------|-------------|
+| `version` | A SemVer requirement — see [SemVer Compatibility](./semver.md) |
+| `registry` | Base URL of the Artifactory registry |
+| `repository` | Repository name within that registry |
 
+## Adding dependencies via the CLI
+
+The `buffrs add` command writes the manifest entry for you:
+
+```bash
+# Caret range (recommended): resolves to the highest 1.x.y
+buffrs add --registry https://my-registry.example.com my-repo/my-lib@^1.0.0
+
+# Exact pin: resolves to exactly 1.2.3
+buffrs add --registry https://my-registry.example.com my-repo/my-lib@=1.2.3
+
+# Latest: omitting the version resolves to the latest available
+buffrs add --registry https://my-registry.example.com my-repo/my-lib
 ```
-buffrs add --registry https://your.registry/artifactory my-repo/my-package@1.2.3
-```
 
-## Local Dependencies
+After adding a dependency, run `buffrs install` to resolve and download it.
 
-Local dependencies are resolved from the local filesystem relative to the
-manifest. They are useful for multi-package repositories where packages depend
-on each other without going through a remote registry.
+## Version requirements
+
+buffrs supports the full range of SemVer requirement operators:
 
 ```toml
-[dependencies]
-my-lib = { path = "../my-lib" }
+version = "^1.0.0"      # >=1.0.0, <2.0.0  (recommended for most deps)
+version = "~1.2.0"      # >=1.2.0, <1.3.0  (patch updates only)
+version = ">=1.5.0"     # any version at or above 1.5.0
+version = "=1.2.3"      # exactly 1.2.3
+version = ">=1.0, <2.0" # explicit intersection
 ```
 
-The `path` field is a relative path from the manifest file to the dependency's
-root directory (the directory containing the dependency's `Proto.toml`).
+The resolver queries the registry and selects the **highest** available version
+satisfying the requirement. See [SemVer Compatibility](./semver.md) for the
+full operator reference.
 
-See [Local Dependencies](../guide/local-dependencies.md) for more information.
+## Local dependencies
 
-## Lockfile
+You can depend on a package in a local directory (useful in monorepos or during
+development):
 
-After adding or modifying dependencies in the manifest, run
-[`buffrs install`](../commands/buffrs-install.md) to resolve and lock them.
-The lockfile (`Proto.lock`) records the exact resolved versions and checksums
-and should be committed to version control.
+```toml
+[dependencies.my-lib]
+path = "../my-lib"
+```
 
-See [Manifest vs Lockfile](../guide/manifest-vs-lockfile.md) for more
-information.
+Local dependencies do not have a version requirement — the package at that
+path is used as-is. They cannot be mixed with a remote entry for the same
+package name.
 
+## The lockfile
+
+Once resolved, the concrete version is recorded in `Proto.lock`. Subsequent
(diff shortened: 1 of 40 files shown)


docs/src/reference/specifying-dependencies.md
```diff
@@ -1,52 +1,81 @@
...
+
+```toml
+[dependencies.my-lib]
+path = "../my-lib"
+```
 
-See [Local Dependencies](../guide/local-dependencies.md) for more information.
+Local dependencies do not have a version requirement — the package at that
+path is used as-is. They cannot be mixed with a remote entry for the same
+package name.
```
LinaUr, line 70: Can we keep the removed lines:
```suggestion
The `path` field is a relative path from the manifest file to the dependency's
root directory (the directory containing the dependency's `Proto.toml`).

Local dependencies do not have a version requirement — the package at that
path is used as-is. They cannot be mixed with a remote entry for the same
package name.

See [Local Dependencies](../guide/local-dependencies.md) for more information.
```

src/resolver.rs
```diff
@@ -51,8 +51,10 @@ pub struct DependencyNode {
     pub source: DependencySource,
     /// Packages that this package depends on
     pub dependencies: Vec<PackageName>,
-    /// Version requirement
+    /// Version requirement from the manifest
     pub version: VersionReq,
+    /// The concrete version that was resolved and downloaded (None for local deps)
+    pub resolved_version: Option<semver::Version>,
```
LinaUr: The file already imports `semver::VersionReq` - import `Version` too:
```
```

```suggestion
    pub resolved_version: Option<Version>,
```

src/lock.rs
```diff
@@ -172,6 +179,11 @@ impl LockedPackage {
         }
     }
 
+    /// Returns true if the locked version satisfies the given requirement
+    pub fn satisfies_requirement(&self, req: &VersionReq) -> bool {
+        req.matches(&self.version)
+    }
```
LinaUr: As far as I can see, this is a dead method.
Consider removing it, or use it in the `find_satisfying` implementations for consistency - those seem to be the intended callers but they use `req.matches(&p.version)` directly.

src/resolver.rs
```diff
@@ -361,76 +364,87 @@ impl<'a> GraphBuilder<'a> {
...
+        } else {
+            let client = Artifactory::new(registry.clone(), self.credentials)
+                .wrap_err_with(|| format!("failed to initialize registry {}", registry))?;
+            self.registry_clients
+                .insert(registry.clone(), client.clone());
+            client
+        };
+
+        // Phase 1: resolve the requirement to a concrete version.
+        // Try the lockfile first (avoids a registry round-trip when the lock is fresh).
```
LinaUr: The existing `tests/cmd/install/workspace/lockfile/stale/mod.rs` test already executes two consecutive installs and verifies the lockfile is reused and extended. However, afaik, it uses exact pins (`=0.1.0`), so it tests the old `Lockfile::get()` path with exact version matching, not explicitly covering the new `find_satisfying()` path with range matching.

Could you add a test for that path please?
For example: run install with `^1.0.0`,  get `1.2.0` locked, run install again, verify it reuses `1.2.0` from the lockfile without hitting the registry

tests/cmd/install/workspace/range_caret_resolution/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+/// Publishes leaf-lib at v1.0.0, v1.1.0, and v1.2.0, then installs with
+/// `^1.0.0`. Expects the lockfile to pin v1.2.0.
+#[test]
+fn fixture() {
+    with_test_registry(|url| {
+        let vfs = VirtualFileSystem::copy(crate::parent_directory!().join("in"));
+        let buffrs_home = vfs.root().join("$HOME");
+        let cwd = vfs.root();
+
+        // Publish leaf-lib at v1.0.0
```
LinaUr, line 15: There seems to be significant code redundancy within and across these tests, like publishing a lib or adding a dependency with semver requirement. Maybe consider writing (shared) helper functions.

src/resolver.rs
LinaUr, line 323: I think there might be a bug here. 
IIUC, when a package is encountered for the second time (via a different path in the dependency tree), the resolver doesn't re-resolve. 

For example, given:
- pkg-1 requires lib-a `^1.0.0`
- pkg-2 requires lib-a `<=1.2.0`
- highest version of lib-a is `1.5.0`

If pkg-1 resolves first, the version is `1.5.0`.
When pkg-2 tries to resolve, `1.5.0` doesn't satisfy `<=1.2.0`, so install fails.
However, in this case I'd expect it to resolve to `1.2.0` across the requirements.

If pkg-2 would resolves first, the version would satisfy pkg-1's requirement.

CMIIW, but since [DependencyMap](https://github.com/helsing-ai/buffrs/blob/main/src/manifest/package.rs#L306) is `HashMap<PackageName, DependencyManifest>`, I believe the dependency iteration order is also non-deterministic. That would make an integration test for the order-dependency inherently flaky.

LinaUr: It's worth adding a test for these cases.

tests/cmd/install/workspace/range_caret_resolution/mod.rs
LinaUr: I created a issue (https://github.com/helsing-ai/buffrs/issues/312) for this, so it's out of scope for this PR.

docs/src/reference/specifying-dependencies.md
heatonmatthew: Good catch.  Fixed.

Please check you're happy.

src/lock.rs
heatonmatthew: Well spotted.  Removed

src/resolver.rs
heatonmatthew: Fixed.

src/resolver.rs
heatonmatthew: Thanks for pointing this out.  It's likely that I'll need to change the resolver algorithm away from the current single-pass recursive descent.  It will need to keep track of aggregated package/requirement information, then fetch recursive transitives and flag any changes implied on already resolved package/requirements.  Then this process would be looped until either 1. all requirements are met or 2. a subset of the requirements are irresolvable.

I'm working on it and will get back to you.

heatonmatthew: My apologies for the delay, I've had an extended illness.  I'll try to get back to finishing this soon.

src/resolver.rs
heatonmatthew: I've now added `tests/cmd/install/workspace/lockfile/range_pin_reuse/mod.rs` to test this path.

src/resolver.rs
heatonmatthew: OK, I've made a bunch more changes and added new tests that should resolve this issue.  Ready for another review, thanks.

heatonmatthew: The `range_cascade_resolution` and `range_downgrade_diamond` tests should exercise this now.

PR #314, opened 2026-05-30
Non-IFD Nix Integration

mara-schulke:
This MR adds a non-IFD variant of the `cache.nix` which circumvents frequent rebuilding and performance bottlenecks on buffrs in nix. Reference on IFD: https://nix.dev/manual/nix/2.26/language/import-from-derivation

### DIFF
--- a/nix/cache-pure.nix
+++ b/nix/cache-pure.nix
@@ -1,0 +1,45 @@
+# Pure-Nix variant of `cache.nix` — same `BUFFRS_CACHE` output, but
+# the URL list is derived by parsing `Proto.lock` with `fromTOML`
+# instead of invoking `buffrs lock print-files` under IFD.
+#
+# Use this when the caller evaluates `vendorDependencies` for a
+# system it can't build (e.g. enumerating `aarch64-darwin` outputs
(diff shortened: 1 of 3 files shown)


nix/cache-pure.nix
```diff
@@ -0,0 +1,45 @@
...
+# Keep this in sync if the on-the-wire layout changes.
+{ fetchurl, runCommand, lib, symlinkJoin, }:
+
+lockfile:
+
+let
+  lock = builtins.fromTOML (builtins.readFile lockfile);
+  packages = lock.packages or [ ];
+
+  fileUrl = pkg:
```
ulucs, line 22: this feels a bit unstable with regards to having to keep the logic in here and inside the rust code in sync at all times. What if we had:
- a buffrs command that downloaded a single dependency (so the hash can be reused for the fixed-output derivation)
- a buffrs command that merges the outputs of above derivations into a cache folder

that way, the whole cache folder logic stays inside the rust binary; and can be changed without breaking things for users

j-baker: maybe we would be better off just having a test that this approach matches the rust behaviour?

PR #317, opened 2026-06-10
Use sets to compare locked dependencies

mara-schulke:
Prevents misleading warnings for users when the locked dependencies are out of order with the downloaded dependencies:

```
:: dependencies mismatch for x@1.0.0: [Qualified { name: PackageName("a"), version: Version { major: 0, minor: 1, patch: 0 } }, Qualified { name: PackageName("b"), version: Version { major: 0, minor: 2, patch: 0 } }] vs [Qualified { name: PackageName("b"), version: Version { major: 0, minor: 2, patch: 0 } }, Qualified { name: PackageName("a"), version: Version { major: 0, minor: 1, patch: 0 } }]. Using first seen.
```

### DIFF
--- a/src/lock.rs
+++ b/src/lock.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 use std::{
-    collections::{BTreeMap, BTreeSet},
+    collections::{BTreeMap, HashSet},
     path::Path,
 };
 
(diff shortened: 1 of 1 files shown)


src/lock.rs
```diff
@@ -12,7 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::{collections::BTreeMap, path::Path};
+use std::{
+    collections::{BTreeMap, BTreeSet},
```
vsiles: For comparison only, we might fall back on HashSet to not pay the full price of BTreeSet's complexity. But that's mostly a nit

vsiles (approved):

PR #319, opened 2026-06-30
fix(package): ignore non-file manifest entries when parsing

Ysh204:
## Summary

Fix package parsing to only consider regular file entries when searching for the package manifest inside a tar archive.

Previously, the parser matched any tar entry whose path ended with the manifest filename, including directories. This could cause parsing to fail if an archive contained a directory named `Proto.toml`.

This change:

* Verifies that the matching tar entry is a regular file using the tar header's entry type.
* Adds a regression test covering an archive that contains both a directory named `Proto.toml` and the actual manifest file, ensuring the parser selects the correct entry.

## Why

This addresses the existing TODO in `Package::parse`. Instead of checking the host filesystem, the parser now uses the tar entry's header to determine whether the matched manifest entry is a regular file, ensuring correct behavior for archives containing non-file entries with the manifest name.

### DIFF
--- a/src/package/compressed.rs
+++ b/src/package/compressed.rs
@@ -193,15 +193,12 @@ impl Package {
             .wrap_err(miette!("corrupted tar package"))?
             .filter_map(|entry| entry.ok())
             .find(|entry| {
-                entry
-                    .path()
-                    .ok()
-                    // TODO(rfink): The following line is a bug since it checks whether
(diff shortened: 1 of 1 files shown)


Ysh204:
I looked into the failing `cargo deny` check and confirmed it also fails on the current `main` branch with the existing `anyhow` (`RUSTSEC-2026-0190`) advisory. This appears to be unrelated to the changes in this PR.

